Issue #10, opened 2020-03-29
External resources CRUD


exbinary:

maebeale:
Considering running the following: any objections/recommendations?  @exbinary @armahillo 

`rails g scaffold ExternalResource name:string website_url:string facebook_url:string phone:string street_address:string city:string state:string zip:string description:string display_on_website:boolean youtube_identifier:string`

exbinary:
Sounds good.

PR #16, opened 2020-03-31
Devise setup - pending specs

maebeale:

PR #19, opened 2020-03-31
Model changes

exbinary:
Some changes to the base models:
* Add `Location` model, associate with `Listing`. Could be useful in #10 . Decided to leave most fields nullable for now to accommodate listings with partial addresses.
* Add `tags` to `Listing` via `acts-as-taggable-array-on` gem.
* Add a `state` enum to `Listing`.
* Rename `Need` -> `Ask`, `Offering` -> `Offer`
* Remove `Resource` model in favor of `ExternalResource` in #10 .

### DIFF
--- a/app/views/listings/_listing.json.jbuilder
+++ b/app/views/listings/_listing.json.jbuilder
@@ -1,2 +1,2 @@
-json.extract! listing, :id, :title, :description, :type, :created_at, :updated_at
+json.extract! listing, :id, :type, :created_at, :updated_at
 json.url listing_url(listing, format: :json)
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,8 @@ gem 'jbuilder', '~> 2.7'
 # Reduces boot times through caching; required in config/boot.rb
 gem 'bootsnap', '>= 1.4.2', require: false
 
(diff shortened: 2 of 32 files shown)


exbinary:
Once #16 is merged, can associate `Listing` with `User`.

app/views/listings/_listing.json.jbuilder
```diff
@@ -1,2 +1,2 @@
-json.extract! listing, :id, :title, :description, :type, :created_at, :updated_at
```
maebeale: I think we'll still need a description? (e.g. $100 cash requested, $50 offered). Relatedly, how will we handle an ask or offer that's not already an available tag? 

Also I think we should consider adding to date by when the listing is needed, and date by when it should be removed from the site.  Do you agree?

exbinary: Yep, we'll need one or two more fields but i was thinking they could be more specifically named than a generic `description`. Maybe not, will see.

I'm working on allowing users to add tags from the form.

I like the idea of `needed_by` and `list_until`. Added an issue to make sure we get to that:
https://github.com/maebeale/mutual-aid/issues/20

PR #25, opened 2020-04-01
Move bootstrap and jquery into webpack

exbinary:
Since rails 6 moved js out of sprockets into webpack, using gems to include them is no longer best practice, and in this case, doesn't seem to work.

Building on #22, used `yarn` to add jquery, popper and bootstrap npm packages and added configuration to get it all working. Note that the bootstrap css is still served via sprockets which is capable of finding the css under `node_modules/`.

A couple of other notes:
* Looks like the `require_tree` and `require_self` sprockets directives were inadvertently removed in 2e5899b. Restored here.
* I'm not sure what `authenticated.scss` is doing; seems to have a lot of custom css that is overriding some bootstrap. ~I've commented it out here and most things still seem to work~. It's still being pulled in via `require_tree` and overriding bootstrap css. We'll need to clean it up i think.
* Alert/notice messages can now be closed.
* The _Toggle Sidebar_ button still doesn't work, but the js snippet in `_sidebar` is firing, so i think this is a bug with the implementation.

At some point, we may consider also moving `font-awesome` out of sprockets into webpack.


app/assets/stylesheets/application.scss
```diff
@@ -1,6 +1,22 @@
...
+ * It is generally better to create a new file per style scope.
+ *
+ *= require_tree .
+ *= require_self
+ */
+
+@import "bootstrap/scss/bootstrap";
 @import "font-awesome";
-@import "./authenticated";
+// @import "./authenticated";
```
maebeale, line 19: we can remove this

maebeale: Instructions had said to convert to scss and remove require tree, etc, but as long as it's working, happy to have it back! Since you put the old syntax back in there (*= vs @ import), should we also change the file extension back to css?

maebeale:
Ok, awesome, @exbinary ! 

I did have bootstrap and font-awesome working this way (followed gemfile instructions), but happy to change to best practice!

Removing unauthenticated only changed one thing, so, yeah, we're good on removing it. I'd stubbed out the sidebar/navbar/page frame with it and hadn't yet verified it could all go.

maebeale:
Going to merge and push to heroku. Thanks again, @exbinary !

app/assets/stylesheets/application.scss
exbinary: Sorry @maebeale, i misunderstood what you had done. I was putting css in `listings.scss`, seeing them not take effect and tracked it down to that change. But i should've realized the appropriate fix would have been to explicitly `import "./listings"` instead of rollback your change.

exbinary:
> Removing unauthenticated only changed one thing, so, yeah, we're good on removing it. I'd stubbed out the sidebar/navbar/page frame with it and hadn't yet verified it could all go.

So because i put `require_tree` back in there, `authenticated` is currently still loading. If i actually remove the file, the sidebar breaks horribly.

I would like to clean up `authenticated.scss` because it overrides a lot of styles globally and it will be hard to know what's coming from bootstrap and what's coming from the styles defined there. LMK if you have ideas for how to reconcile. Down to talk it through if that's easier.

maebeale:
Yeah, I def didn't want to leave it in there either, @exbinary. I can work on that tonight.

app/assets/stylesheets/application.scss
maebeale: oh, no worries!!! :) but thanks for commenting that, @exbinary !
i think we do need to change the extension if going w require_tree tho ?

exbinary: Hmm, i didn't know that was a constraint. It seems to be getting imported doesn't it?

maebeale: Seems so! That makes sense, that the *= css syntax would be ok in scss file...so, yeah, prob fine to leave as scss.

PR #28, opened 2020-04-01
Add name email phone to form

exbinary:
Still need to add contact_preference

Also needs more validation and some more thinking about the downsides of
making email/phone required.


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_04_01_072342) do
+ActiveRecord::Schema.define(version: 2020_04_01_123807) do
```
maebeale, line 13: this is my fault. sending a PR to get this checked in correctly.

PR #29, opened 2020-04-04
Switch to Bulma

exbinary:
Rips out Bootstrap in favor of Bulma.

@Ry-Dowd, i've commented out most other css throughout the app so you have close to completely clean slate. The Bulma css should be available site-wide but the only place any of the classes are actually used so far is on `listings/new`, to replace some of the bootstap that used to be on that page. There are still a number of pages with bootstrap classes in the markup, but they shouldn't have any effect; we can clean them up as we touch those templates.

This makes all existing pages pretty unpresentable 😅 .

Also yanks the sidebar and navbar in anticipation of a new Bulma+Vue layout. @maebeale, i couldn't think of an easy way to set Ryan up and retain all the work you put into the layout. Definitely open to a different approach. Sorry!


maebeale:
This is great, @exbinary !!!!!

exbinary:
Merging into release branch `v0.2`.

PR #30, opened 2020-04-05
Integrate Vue

exbinary:
Adds Vue and configures it for use in a hybrid app:
* Configure webpack to expose JS snippets in `entry_points/` to the browser.
* These can then be invoked in `<script>` elements, with params, from server rendered pages.
* Entry point snippets will then import and instantiate individual Vue instances, passing along any props from the page.

Also disable turbolinks.


Ry-Dowd:
Looks good! I'll catch up with you this afternoon for a walkthrough on the integration.

PR #31, opened 2020-04-08
fix failing tests with the addition of Faker

h-m-m:
I added Faker and fixed the tests that `bin/rspec` said were failing. I'm okay with editing this change to not use faker, if anyone prefers that.

Here are some decisions this commit goes along with that we might change soon based on new requirements:
- we're not accepting zip+4 format yet
- we're not accepting Canadian postal codes yet
- listings currently must have both an email and a phone number

### DIFF
--- a/spec/requests/listings_spec.rb
+++ b/spec/requests/listings_spec.rb
@@ -3,7 +3,10 @@ require 'rails_helper'
 RSpec.describe "/listings", type: :request do
   let(:valid_attributes) {{
     location_attributes: {zip: "12345"},
-    tags: ["", "cash"]
+    tags: ["", "cash"],
+    name: Faker::Name.name,
+    email: Faker::Internet.email,
+    phone: Faker::PhoneNumber.phone_number
   }}
 
   let(:invalid_attributes) {{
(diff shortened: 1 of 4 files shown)


spec/requests/listings_spec.rb
```diff
@@ -83,19 +86,20 @@
...
 
   describe "PATCH /update" do
-    let(:listing) { create(:listing) }
+    let(:listing) { create(:listing, :with_contact_info) }
 
     context "with valid parameters" do
+      let(:new_street_address) { Faker::Address.street_address }
       let(:new_attributes) {{
-        location_attributes: {street_address: 'changed'},
+        location_attributes: {street_address: new_street_address, zip: Faker::Address.zip(state_abbreviation: 'MI')},
```
h-m-m, line 94: Passing `state_abbreviation` here is a kludge to coerce Faker into not trying zip+4 values

I was going to make the form allow zip+4, using the regex from https://github.com/dgilperez/validates_zipcode/blob/6345983cb72dd3a2776ca06a42039be7df75d307/lib/validates_zipcode/cldr_regex_collection.rb#L15. The test still failed due to DB limitations, so this is something we can revisit later

PR #32, opened 2020-04-08
Add Circle CI

h-m-m:
Here's a Circle CI config with relevant config changes. Important notes:
- I added the `dotenv` gem because that will help with keeping testing, development, and deployment variables consistent. It may also be helpful when it comes to welcoming new people with unusual computer configurations who want to contribute
- This includes the changes in https://github.com/maebeale/mutual-aid/pull/31 because otherwise the tests would fail and because rebasing without those test fixes and then rebasing again after that's merged in would be annoying and a little silly. That said, I put the tests in a separate PR to make it easy for everyone to look at them separately. If you want to review code changes, you should probably review https://github.com/maebeale/mutual-aid/pull/31 first

Otherwise, it's not too unusual as CircleCI configs go

Once this is merged in, using the default options on the CircleCI side to tell it to run this config should work just fine

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ commands:
             set -euxo pipefail
             bundle config path .bundle
             bin/rails db:setup
-            bin/rspec
+            bin/rspec --format progress --format RspecJunitFormatter -o ~/rspec/rspec.xml
 
 jobs:
   bundle:
@@ -72,6 +72,8 @@ jobs:
     steps:
       - checkout
       - test_suite
+      - store_test_results:
+          path: ~/rspec
 
 workflows:
   version: 2.1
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,87 @@
...
+      - <<: *docker_image
+    steps:
+      - checkout
+      - yarn_install
+  test:
+    docker:
+      - <<: *docker_image
+        environment:
+          RAILS_ENV: test
+      - image: circleci/postgres:latest
```
h-m-m, line 67: Once this is merged in and set up in CircleCI, I'd like to pin the postgres version and abstract that up to the top of the file like I did with the container for the Ruby version

PR #55, opened 2020-04-14
Change resources submission process to be public-facing w a separate admin workflow

maebeale:
This starts to move admin workflows into an admin controller, and open endpoints for the public (e.g. to view CommunityResources). It's touching a lot of things, so we _may_ want to consider breaking it apart?


app/assets/stylesheets/application.scss
```diff
@@ -14,22 +14,31 @@
  *= require_self
  */
 
+$input-placeholder-color: hsl(0, 0%, 86%); //has-text-grey-lighter
```
maebeale, line 17: These changes are in another PR, but I pulled them in here bc I'm building upon the styling to make the public-facing pages (and the internal ones, but...).

app/controllers/application_controller.rb
```diff
@@ -2,5 +2,10 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery with: :exception
   before_action :authenticate_user!
+  before_action :admin_status
```
maebeale, line 5: Not sure how y'all are in to application-level callbacks, but in apps where there are shared partials, I find it SUPER helpful. It's also a quick way to toggle between admin and public viewing. We're going to end up with more permission levels than just admin, too.

app/controllers/external_resources_controller.rb
```diff
@@ -22,7 +22,7 @@ def create
 
     respond_to do |format|
       if @external_resource.save
-        format.html { redirect_to external_resources_path, notice: 'External resource was successfully created.' }
+        format.html { redirect_to @admin_status ? external_resources_path : community_resources_path, notice: "External resource was successfully submitted.#{ " We'll review." unless @admin_status }" }
```
maebeale, line 25: Here's an example of why having @admin_status can be so useful. After an admin creates a resource, they'd go to the admin-index where there's a condensed index view w edit links, whereas a non-authenticated user should get taken back to the pretty list. :)

app/controllers/external_resources_controller.rb
```diff
@@ -34,7 +34,7 @@ def create
   def update
     respond_to do |format|
       if @external_resource.update(external_resource_params)
-        format.html { redirect_to external_resources_path, notice: 'External resource was successfully updated.' }
+        format.html { redirect_to community_resources_path, notice: 'External resource was successfully updated.' }
```
maebeale, line 37: The PublicController endpoint is called community_resources, _and_, I think we should rename ExternalResource to CommunityResource in general. In that case we'd want to rename this route to something like `public_community_resources_path` ?

app/controllers/external_resources_controller.rb
```diff
@@ -59,8 +59,17 @@ def set_external_resource
 
     # Only allow a list of trusted parameters through.
     def external_resource_params
-      params.require(:external_resource).permit(:name, :website_url, :facebook_url,
-                                                :phone, :description, :display_on_website,
-                                                :youtube_identifier, :location_id, tags: [])
+      params.require(:external_resource).permit(
+          :description,
+          :display_on_website_end,
+          :display_on_website_start,
```
maebeale, line 65: Converting display_on_website from boolean to date range. (I know postgres supports date range fields, but leaving them separate fields allows the user to leave these blank and therefore not part of what hides/shows things on the website.)

app/controllers/external_resources_controller.rb
```diff
@@ -59,8 +59,17 @@ def set_external_resource
...
+          :facebook_url,
+          :location_id,
+          :name,
+          :phone,
+          :reviewed,
```
maebeale, line 70: Added the boolean  `reviewed` to help with workflows -- Admins can log in and see how many "unreviewed" resources there are...

app/controllers/external_resources_controller.rb
```diff
@@ -59,8 +59,17 @@ def set_external_resource
...
+          :website_url,
+          :youtube_identifier,
```
maebeale, line 72: This is so someone could embed a video on resource show.

app/controllers/public_controller.rb
```diff
@@ -5,4 +5,8 @@ class PublicController < ActionController::Base
   def landing_page
   end
 
+  def community_resources
+    @external_resources = ExternalResource.all
```
maebeale, line 9: This needs to get changed to a filtered list rather than `all`

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,7 @@
 module ApplicationHelper
+
+  def yes_no(boolean)
+    "<span class='#{ boolean ? "fa fa-check-circle has-text-success" : "fa fa-ban" }'></span>".html_safe
+  end
```
maebeale, line 5: This is a utility I find very useful, but not sure if it's an antipattern or something y'all would rather not use? (And, if ok to use, ok to include in this PR where I use it, or should these always be extracted?)

app/javascript/components/Orientation.vue
```diff
@@ -22,7 +22,10 @@
       </div>
       <div class='column'>
         <div>
-          <h2 class='subtitle is-4'><a href='/external_resources/new'>Is there an outside resource we should know about?</a></h2>
```
maebeale, line 25: Changed this from one sentence to match the format of the others and go directly to the `new_external_resource` endpoint.

app/models/user.rb
```diff
@@ -4,4 +4,8 @@ class User < ApplicationRecord
          :recoverable, :rememberable, :validatable,
          :confirmable, :lockable, :timeoutable, :trackable# and :omniauthable
 
+  def admin?
+    true # TODO - connect this to pundit
+  end
```
maebeale, line 9: this will need some Pundit implementation, but making anyone who logs in an admin atm. Is this ok w y'all?

app/views/devise/registrations/new.html.erb
```diff
@@ -5,8 +5,8 @@
   <div class="col-md-4 text-center">
     <% if current_user %>
       <h2>You are already logged in</h2>
-      <% if current_user&.admin? %>
-        <%= link_to("<span class='fa fa-menu #{ENV['NAVBAR_ICON_COLOR_CLASS']}'></span><span class='text-black text-bold'> Go back to admin</span>".html_safe,
+      <% if @admin_status %>
+        <%= link_to("<span class='fa fa-menu #{ENV['NAVBAR_ICON_COLOR_CLASS']}'></span><span class='has-text-black text-bold'> Go back to admin</span>".html_safe,
```
maebeale, line 9: changed to bulma class!

app/views/external_resources/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= simple_form_for(external_resource) do |form| %>
+<%= simple_form_for @external_resource, wrapper: :default, defaults: { wrapper: false } do |form| %>
```
maebeale, line 1: This calls the default section of the `simple_form_bulma file`

app/views/layouts/_navbar.html.erb
```diff
@@ -18,7 +18,7 @@
           <a href="/listings">Listings</a>
         </div>
         <div class="navbar-item">
-          <a href="/external_resources">External Resources</a>
+          <a href="/community_resources">Community Resources</a>
```
maebeale, line 21: Change navbar to go to the public community_resources vs the admin external_resources

config/initializers/simple_form.rb
```diff
@@ -0,0 +1,190 @@
+# Use this setup block to configure all options available in SimpleForm.
```
maebeale, line 1: these files are in another PR, but I merged that branch in so I could streamline forms above

config/routes.rb
```diff
@@ -5,7 +5,7 @@
   resources :listings
 
   resources :public, only: [:landing_page]
-
+  get "/community_resources", to: "public#community_resources"
```
maebeale, line 8: Do y'all prefer `resources...only: []` to `get`s like this?

spec/factories/external_resources.rb
```diff
@@ -1,12 +1,7 @@
 FactoryBot.define do
   factory :external_resource do
     name { "MyString" }
-    website_url { "MyString" }
```
maebeale, line 4: Remove some of these non-required fields. The scaffolds add all fields by default.

spec/views/external_resources/edit.html.erb_spec.rb
```diff
@@ -27,13 +29,22 @@
 
       assert_select "input[name=?]", "external_resource[phone]"
 
-      assert_select "input[name=?]", "external_resource[description]"
+      assert_select "textarea[name=?]", "external_resource[description]"
 
-      assert_select "input[name=?]", "external_resource[display_on_website]"
+      # binding.pry # TODO - helppppppppp
```
maebeale, line 34: Having trouble figuring out how to `assert_select` the date part inputs. Would LOVE help on this!

app/controllers/application_controller.rb
h-m-m: Is it time to consider moving this into pundit?

spec/views/external_resources/edit.html.erb_spec.rb
```diff
@@ -8,17 +8,19 @@
...
     ))
   end
 
   it "renders the edit external_resource form" do
+    puts render
     render
-
     assert_select "form[action=?][method=?]", external_resource_path(@external_resource), "post" do
-
+      # binding.pry
```
maebeale, line 23: leaving this here until tests pass

PR #67, opened 2020-04-16
[DRAFT] Add translatable

maebeale:
Add Mobility gem for translation support (details below)
TODO: 
- add UI elements for adding translations
- refactor list of locales
- adjust serializer (once it exists) with translations

* Mobility gives us the following options (per https://dejimata.com/2017/3/3/translating-with-mobility):
`resource.attributename_[locale abbreviation] = "foo"`
`resource.save!`

* Mobility, using key_value backend strategy, stores translations in one of two polymorphic tables, string_translations & text_translations, which become association methods after extending the module and including the "translates" for attributes needing translation

* For translations on a Post model:
`class Post
  extend Mobility
  translates :title
  translates :content, type: :text
end`

* Accessing all translations for a resource/record/instance
`resource.string_translations` (table where most translations will be stored)
`resource.text_translations`

* To change locales in the console:
`I18n.default_locale` # defaults to :en
`Mobility.locale = :es` # to switch it temporarily

* Other usage examples:
```
Post.find_by(title_en: "Mobility")
#=> #<Post:0x... id: 123 content_en: "Mobility" content_ja: "モビリティ"...>
Post.find_by(title_ja: "モビリティ")
#=> #<Post:0x... id: 123 content_en: "Mobility" content_ja: "モビリティ"...>

post = Post.first
translation = post.translations.find do |translation|
  translation.locale == "en"
end
#=> #<PostTranslation:0x... id: 123 locale: "en" ...>
translation.title
#=> "Translating with Mobility"
translation.content
#=> "In the first real post on this blog..."

post = Post.first
post.text_translations.find do |translation|
  translation.key == "title" && translation.value == "Translating with Mobility"
end
#=> #<TextTranslation:0x...
      id: 123
      locale: "en"
      key="title"
      value="Translating with Mobility"
      ... >```

### CODE REVIEW

**Comment by armahillo** [comment]
What are the advantages of using the `Translatable` gem over using native `i18n` support and having the locale YMLs loaded based on user-selected locale?

**Comment by maebeale** [comment]
Hey there, @armahillo !!! VERY open to your suggestions on this, but here's the thinking below.

* We won't know the data needed for the YML at deploy bc it's translations of other user-created data
* By having it be data that is structured and accessible in the same way as the other system-level datasets (SystemTags, SystemLocations) any of their admins can configure the system and make changes on the fly through the UI.
* We will need to be able to search on the translated data, so having it in SQL along w the english versions is ideal.

**Comment by exbinary** [comment]
@maebeale, i definitely hear the need for dynamic, user created translations. What do you think about pulling in [`Mobility`](https://github.com/shioyama/mobility), a gem designed to support this use case? I think it would do a lot of the heavy lifting for us. There's also [`Globalize`](https://github.com/globalize/globalize), an older, more widely used gem, but my cursory review of both has me leaning towards Mobility.

**Comment by maebeale** [comment]
YESSSSS, @exbinary!!!!

**Comment by maebeale** [comment]
I can change this pr to remove that table

File: app/controllers/application_controller.rb [hunk 1]
```diff
@@ -3,9 +3,19 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   before_action :authenticate_user!
   before_action :admin_status
+  around_action :switch_locale
```
maebeale, line 6: This is how to rig up switching desired locale by using url params

app/controllers/application_controller.rb
```diff
@@ -3,9 +3,19 @@ class ApplicationController < ActionController::Base
...
 
   def admin_status
     @admin_status = params[:admin] ? YAML.load(params[:admin]) : current_user&.admin? # allows admin user to simulate with param=false
   end
 
+  def switch_locale(&action)
+    locale = params[:locale] || current_user.try(:locale) || I18n.default_locale
```
maebeale, line 13: Ok to add the field :locale to User so we can set a default for them?

app/models/announcement.rb
```diff
@@ -1,4 +1,7 @@
 class Announcement < ApplicationRecord
+  extend Mobility
+  translates :name
+  translates :description
```
maebeale, line 4: An alternative is to set the type explicitly: `translates :description, type: :text`

Not sure if we care about splitting these. Maybe this is to help w querying speed later?

app/models/community_resource.rb
```diff
@@ -1,5 +1,8 @@
 class CommunityResource < ApplicationRecord
   taggable_array :tags
+  extend Mobility
```
maebeale, line 3: Ok to put this below taggable_array declaration?

app/models/custom_form_question.rb
```diff
@@ -1,4 +1,8 @@
 class CustomFormQuestion < ApplicationRecord
+  extend Mobility
+  translates :name
+  translates :option_list, type: :string # if this is json, ok to say string here?
```
maebeale, line 4: Wasn't sure how best to support translations of a nested hash/json/etc. Any suggestions?

config/application.rb
```diff
@@ -37,5 +37,12 @@ class Application < Rails::Application
       g.jbuilder          false
       g.scaffold_stylesheet false
     end
+
+    # Permitted locales available for the application
+    I18n.available_locales = [:ar, :cs, :en, :es, :de, :fr, :pt, :th]
```
maebeale, line 42: Would love to pull this list from the one committed in the yml rather than this list, but wasn't sure how. Any suggestions? And/or would you prefer to limit language support list? (I wouldn't, esp for communities that have large immigrant populations.)

config/application.rb
```diff
@@ -37,5 +37,12 @@ class Application < Rails::Application
...
+    config.i18n.enforce_available_locales
```
maebeale, line 43: If we _do_ go w available_locales, we should enforce them. Otherwise prob don't need this?

config/application.rb
```diff
@@ -37,5 +37,12 @@ class Application < Rails::Application
...
+    # Mobility gem will fallbacks to locale set here
+    I18n.default_locale = :en
+    config.i18n.fallbacks = true
```
maebeale, line 46: This allows Mobility to pull the value for the default_locale (announcement.name, even when announcement.name_fr is blank).

config/initializers/mobility.rb
```diff
@@ -0,0 +1,135 @@
+Mobility.configure do |config|
+  # Sets the default backend to use in models. This can be overridden in models
+  # by passing +backend: ...+ to +translates+.
+  config.default_backend = :key_value
```
maebeale, line 4: This is for the polymorphic strategy -- the one I'd planned before adding this gem, and the one the gem docs recommend.

config/initializers/mobility.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  # By default, Mobility uses the +translates+ class method in models to
+  # describe translated attributes, but you can configure this method to be
+  # whatever you like. This may be useful if using Mobility alongside another
+  # translation gem which uses the same method name.
+  config.accessor_method = :translates
```
maebeale, line 10: Set the keyword for the line to add to models

config/initializers/mobility.rb
```diff
@@ -0,0 +1,135 @@
...
+  # (see below).
+  #
+  # config.default_options[:dirty] = true
+
+  # No fallbacks are used by default. To define default fallbacks, uncomment
+  # and set the default fallback option value here. A "true" value will use
+  # whatever is defined by +I18n.fallbacks+ (if defined), or alternatively will
+  # fallback to your +I18n.default_locale+.
+  #
+  config.default_options[:fallbacks] = true
```
maebeale, line 51: Enable fallbacks

config/initializers/mobility.rb
```diff
@@ -0,0 +1,135 @@
...
+  # don't want to do that).
+  #
+  # config.default_options[:default] = ...
+
+  # Uncomment to enable locale_accessors by default on models. A true value
+  # will use the locales defined either in
+  # Rails.application.config.i18n.available_locales or I18n.available_locales.
+  # If you want something else, pass an array of locales instead.
+  #
+  config.default_options[:locale_accessors] = true
```
maebeale, line 69: Gives us nice methods like `announcement.name_fr = 'french name'`

config/initializers/mobility.rb
```diff
@@ -0,0 +1,135 @@
...
+  #
+  # config.default_options[:fallthrough_accessors] = true
+
+  # You can also include backend-specific default options. For example, if you
+  # want to default to using the text-type translation table with the KeyValue
+  # backend, you can set that as a default by uncommenting this line, or change
+  # it to :string to default to the string-type translation table instead. (For
+  # other backends, this option is ignored.)
+  #
+  config.default_options[:type] = :string
```
maebeale, line 89: This makes us not have to declare type for every `translates` declaration in models

config/initializers/mobility.rb
```diff
@@ -0,0 +1,135 @@
...
+end
+
+### Mobility gives us the following options (per https://dejimata.com/2017/3/3/translating-with-mobility):
```
maebeale, line 92: Some helpful snippets I found from the various docs. Also pasted this into the PR message.

app/models/announcement.rb
exbinary: I'm seeing [this note](https://github.com/shioyama/mobility#getting-started) in the docs:
> Note: When using the KeyValue backend, use the options hash to pass each attribute's type:
> This is important because this is how Mobility knows to which of the two translation tables it should save your translation.

Although i'm a bit surprised Mobility doesn't just infer this from the column type 🤔 .

app/models/community_resource.rb
exbinary: I think we're removing `taggable_array` soon. I do think there's a soft norm of `extend` and `include` coming first in a class declaration before anything else?

app/models/custom_form_question.rb
exbinary: That sounds fine, but what does `option_list` hold again? It might need to be `:text` if the json could be on the longer side?

config/initializers/mobility.rb
maebeale: @exbinary i saw the note in the docs (that you quoted in a difft comment) re key_value needing to know which table, but it seemed that adding this was correctly making things go to the string_translations table ?

app/models/announcement.rb
maebeale: right. but if we tell it the default is :string, it seems to work that we only need to explicitly state :text if we do NOT want to store it in the string_translations table.

@exbinary this comment was trying to get at asking -- why would we ever want to store something in the text_translations table? my thinking was maybe to have longer strings in a table together so the string table is as fast as it can be? using both string_translations and text_translations doubles the joins, so i wasn't clear on why we would ever choose anything other than the default (string_translations)

app/models/community_resource.rb
maebeale: yup re removing. yeah, i thought that too re extend and include, but wasn't sure if y'all had preferences re the taggable_array since it is a kind of inclusion.

app/models/custom_form_question.rb
maebeale: If someone adds a select or multiselect input_type, they'll need to provide this option_list. i don't forsee that being super long, but it's definitely possible.

It isn't clear that the string_translations table could store any json, and that's what I was trying to get at. Wouldn't we have to change the string_translations migration for that @exbinary ? Or does it "just work" TM ?

config/initializers/mobility.rb
exbinary: Oh i see, so we'd only have to override this for `:text` columns? I like that.

app/models/announcement.rb
exbinary: Oh, i see what you were asking. Sorry, i misunderstood 🤦.

That's a great question.

AFAIC see, the only difference between the two tables is the type of the `value` column, string vs. text.

From https://www.postgresql.org/docs/9.1/datatype-character.html
> Tip: There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs. In most situations text or character varying should be used instead.

Makes me think text is superior and there's no reason to use the string table?

app/models/custom_form_question.rb
exbinary: Another great question!

I don't think this will 'just work', not fully.

I'm guessing it'll pull the data out fine, since its string-like, but we won't be able to query against it with any type-specific operators. So eg, we wouldn't be able to use any array or json querying operators. Worth testing it out, probably.

PR #68, opened 2020-04-18
[REFERENCE] playing around w fields for ask vs offer forms

maebeale:
@exbinary these files might be a good starter reference?

for the checkbox section on Asks, maybe it could have an accordion that opens if you select it, and then you get the other fields: Urgency dropdown, Date Requested, Notes, [is_ongoing] boolean.

for checkbox section on Offers,  instead of Date Requested or Ugency, maybe two subfields: 
* something about # times per week (or # times total?)
* Notes 
?

Re order of fields:
* For Asks, we prob should put our offerings as first thing they see, rather than asking contact info asap, so the order was swapped for Ask vs Offer (hence the shared partials)

Still lots of nuances to think about here re creating accounts and letting people have some preferences re what will/will not be public. 

The fields in these draft forms aren't connected to the models or current db -- just was spitballing to talk w folks.

We'll also need to create Locations w the addy data (that's connected to a SystemLocation ?)

Screen Shot 2020-04-17 at 9 41 32 AM
Screen Shot 2020-04-17 at 9 41 49 AM
Screen Shot 2020-04-17 at 9 42 02 AM
Screen Shot 2020-04-17 at 9 42 08 AM

### DIFF
--- a/app/views/listings/_ask_form.html.erb
+++ b/app/views/listings/_ask_form.html.erb
@@ -1,0 +1,52 @@
+<% ask ||= listing %>
+<div class=''>
+  <%= form_with(model: listing, local: true, class: 'listing-form') do |form| %>
+    <% if listing.errors.any? %>
+      <div id="error_explanation">
+        <h2 class="subtitle"><%= pluralize(listing.errors.count, "error") %> prohibited this listing from being saved:</h2>
+        <ul>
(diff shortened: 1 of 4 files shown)


maebeale:
@exbinary ok to close this branch?

exbinary:
Yep, this is all in Vue now.

PR #69, opened 2020-04-22
Add js testing with mochapack

h-m-m:
### Description

This adds JS testing. 

This expects all JS test files to live under the folder `test` and end with `.spec.js`. This is a convention I've seen in other projects. Honestly, I don't like this structure. I feel like having a `spec` folder and a `test` folder is weird and confusing. I'm not sure what to use instead, however.

This will allow you to run JS tests by:

* running `npm test` or `yarn test`, which will run the JS tests only
* running `bin/test` which will run the JS tests after running the rspec tests and only if the rspec tests succeed
* triggering a CircleCI run. The CircleCI uses `yarn test` instead of `npm test` because yarn handles passing through and appending additional arguments nicely

### Notes

There is probably a way to write `config/webpack/test.js` so that the test run more efficiently. I'm not up enough on webpack to understand that yet.

I made some "peer dependencies" explicit in the package.json file. I'm unsure how important this was.

### DIFF
--- a/config/webpack/test.js
+++ b/config/webpack/test.js
@@ -1,5 +1,6 @@
 process.env.NODE_ENV = process.env.NODE_ENV || 'development'
 
 const environment = require('./environment')
-
+var babelLoader = environment.loaders.get('babel')
+babelLoader.include.push(babelLoader.include[0].concat('../../test'))
 module.exports = environment.toWebpackConfig()
--- a/test/setup.js
+++ b/test/setup.js
@@ -1,0 +1,1 @@
+require('jsdom-global')()
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,35 +16,49 @@ commands:
           key: bundler-cache-{{ checksum ".ruby-version" }}-{{ checksum "Gemfile.lock" }}
           paths:
             - .bundle
-  yarn_install:
-    description: "Performs node module installation with yarn, relying on the CircleCI cache for performance"
+  node_modules_install:
+    description: "Performs node module installation, relying on the CircleCI cache for performance"
     steps:
       - restore_cache:
           keys:
-            - yarn-cache-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}
+            - node-cache-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}-{{ checksum "package-lock.json" }}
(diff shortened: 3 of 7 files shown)


config/webpack/test.js
```diff
@@ -1,5 +1,6 @@
 process.env.NODE_ENV = process.env.NODE_ENV || 'development'
 
 const environment = require('./environment')
-
+var babelLoader = environment.loaders.get('babel')
+babelLoader.include.push(babelLoader.include[0].concat('../../test'))
```
h-m-m, line 5: If we want to change the location of the tests, editing this line in addition to the relevant `scripts` line in the package.json will be enough to do it

db/schema.rb
```diff
@@ -10,11 +10,41 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_04_12_132525) do
+ActiveRecord::Schema.define(version: 2020_04_15_020109) do
```
exbinary, line 13: Guessing you don't want these changes on this branch?

test/setup.js
```diff
@@ -0,0 +1 @@
+require('jsdom-global')()
```
exbinary, line 1: What does this do?

db/schema.rb
h-m-m: Yup, that was a `git add -a` accident

test/setup.js
h-m-m: This loads https://github.com/jsdom/jsdom so that you can simulate a fairly standards-compliant DOM in your tests without needing to run a browser. There's a note [in the Vue documentation](https://vue-test-utils.vuejs.org/guides/choosing-a-test-runner.html) that something that can behave like this is required for the Vue Test Utils to work. Without the test utils, we don't get the `mount` and `shallowMount` methods for pulling individual Vue components into a test.

We're not actually using this yet, though, so I'm fine with leaving this line out until our first test that requires it

h-m-m: Leaving a reference to this example test generator from Vue so that I can easily find it later: https://github.com/vuejs/vue-cli/blob/bd7ad93cae9711ca30b05c31c185e1ce56bcb1de/packages/%40vue/cli-plugin-unit-mocha/setup.js#L1

PR #71, opened 2020-04-23
Dependency tweaks

exbinary:
Two 'fixes':
1. Remove `package-lock.json` in favor of `yarn.lock` (the latter warns against maintaining both).
2. Add explicit devDependency on babel-loader. Not really sure why this was needed, but `yarn test` was failing w/o it. It was already in `yarn.lock` as a dependency of `webpack`, but for some reason wasn't being picked up. Tried a fresh `yarn install` after blowing away `/node_modules`, to no avail.


h-m-m:
Oh! I was having the babel-loader problem intermittently, and assumed it was something I was doing wrong. It often went away after I fixed a mistake in what I was working on and/or restarted webpack-dev-server, and it never showed up in Circle. Very happy to see this fixed.

I'll add a commit here that cleans up the npm & package-lock.json references in the Circle config

exbinary:
> I'll add a commit here that cleans up the npm & package-lock.json references in the Circle config
Oh sweet, thanks for fixing that -- i saw the failure on CCI but wasn't sure how to fix it.
Merging.

PR #72, opened 2020-04-23
Add prettier config

exbinary:
Definitely open to discussion on any of these settings and am not tied to any of them.

Does _not_ integrate any automated checking or fixing for now. Wanted to give us time to agree on conventions and try them out before enforcing them in any automated way. Also allows us to backburner this PR.

See https://prettier.io/docs/en/editors.html for ways to integrate into editors.


maebeale:
Love this plan -- to get it working well before enforcing!

h-m-m:
I am also 100% with this

PR #73, opened 2020-04-23
Allow `yarn test` to run alongside `bin/webpack-dev-server`

exbinary:
Currently, running tests will break `webpack-dev-server` because the testing build chain writes to the same `manifest.json` file used by the dev-server, but with incompatible pack paths.

This configures the test env to not write the manifest to disk.

I'm not sure if there are other unintended side-effects to this change. An alternative if this doesn't pan out is to instead change the `publicPath` the manifest is written out to.


config/webpack/test.js
```diff
@@ -1,6 +1,15 @@
-process.env.NODE_ENV = process.env.NODE_ENV || 'development'
+process.env.NODE_ENV = process.env.NODE_ENV || 'test'
```
exbinary, line 1: Unrelated to the manifest fix, but seems like a better default here, yeah?

h-m-m: Oh, wow! This is not as unrelated to the manifest fix as you think!

I saw that there's a `packs` and `packs-test` folder under `public`, and assumed that folder difference existed to prevent this problem. I guess that should be coming from https://github.com/maebeale/mutual-aid/blob/7175f842d5d99fc021bfb4ffa016a8b2fa36fa2c/config/webpacker.yml#L80-L85

So after looking over your changes here, I read up on the webpacker documentation and tried testing a few things. Here's what I found:

First, for some reason the environment variable I was trying to set in https://github.com/maebeale/mutual-aid/blob/7175f842d5d99fc021bfb4ffa016a8b2fa36fa2c/package.json#L25-L27 is nowhere to be seen when the test.js file gets evaluated.

Second, if you add one more line to the test.js file at the top that is 
```javascript
process.env.RAILS_ENV = process.env.RAILS_ENV || 'test'
```

It will trigger webpacker to pull in the proper line from the webpacker.yml and that alone should be enough to fix the core problem here

h-m-m: I like it. My only suggestion is to reorder it to not bury the lede. Better to start with saying this line allows you run dev and test side-by-side without additional config, then have the explanation as to why

Also, this change makes the problem you've fixed in https://github.com/maebeale/mutual-aid/pull/71 happen reliably in CI, so that's super interesting. Thankfully, it's happening only because of they are separate PRs

exbinary: > Also, this change makes the problem you've fixed in #71 happen reliably in CI, so that's super interesting. Thankfully, it's happening only because they are separate PRs

Interesting!

Ok to merge?

maebeale: Seems so??? lol

PR #77, opened 2020-04-26
Remove public layout since it's identical to layouts/application

exbinary:
What it says ^^ 😁


maebeale:
does this def work tho? i thought it was needed by the difft controllers bc both inherit from `ActionController::Base` (unlike the rest of the rails controllers that inherit from `ApplicationController` that uses `application` layout) ?

also, there was a slight variation in that the application layout had a div around the yield, but i'd rather it had that wrapper anyway, so, if this defaults, then, great. 

originally i was thinking the public_controller wouldn't even have the navbar, so def would be a difft layout (to serve things like the intake forms without a navbar so they feel less like part of a system), but rn, yeah, they both have the navbar...

maebeale:
relatedly, do you still want to do away with the `public_controller` and `admin_controller` ?
they help me as far as my mental model, but open to your suggestions.

exbinary:
> does this def work tho @exbinary ? i thought it was needed by the difft controllers bc both inherit from `ActionController::Base` (unlike the rest of the rails controllers that inherit from `ApplicationController` that uses `application` layout) ?

On current master, the only controller not inheriting from `ApplicationController` is `PublicController`. Not sure which other controller you're referring to?

`PublicController` is defaulting to `layouts/public`, i believe by matching on the controller name. On this branch, i set the layout explicitly in `PublicController`.

> also, there was a slight variation in that the application layout had a div around the yield, but i'd rather it had that wrapper anyway, so, if this defaults, then, great.

I wasn't seeing a diff on master, but sounds like that's not an issue either way.

> originally i was thinking the public_controller wouldn't even have the navbar, so def would be a difft layout (to serve things like the intake forms without a navbar so they feel less like part of a system), but rn, yeah, they both have the navbar...

I do think we'll eventually need a different layout. I just don't like maintaining two identical ones because its so easy for them to get inadvertently out of sync and then its hard to know what was an intentional deviation. So i'd prefer to extract one when we need it.

But if they've already deviated on your scaffold branch, i can just dump this branch.

exbinary:
> relatedly, do you still want to do away with the `public_controller` and `admin_controller` ?
> they help me as far as my mental model, but open to your suggestions.

What seems most straightforward to me is to have both a `PublicController` and `AdminController` and have all controllers inherit from one or the other.

But our needs might be a bit more complicated than that: we will likely have controllers with some public actions and some admin-only actions. In those cases, i'm guessing we'll rely on Pundit to guard actions based on roles? I really don't like `skip_before_action` directives!

Maybe it makes sense for `ApplicationController` to be the 'public' one and have `AdminController` be the superclass of any resources that are exclusively admin-only?

h-m-m:
> 
> But our needs might be a bit more complicated than that: we will likely have controllers with some public actions and some admin-only actions. In those cases, i'm guessing we'll rely on Pundit to guard actions based on roles? I really don't like `skip_before_action` directives!
> 
> Maybe it makes sense for `ApplicationController` to be the 'public' one and have `AdminController` be the superclass of any resources that are exclusively admin-only?

I like all of this as an approach. I think we know enough that the `AdminController` has a clear use, and then we can wait to reorganize the rest if they become equally clear, too

PR #78, opened 2020-04-27
Pin Bulma version to Buefy, scope overrides for SimpleForm

exbinary:
Two tweaks became apparent as i started to use Buefy:
1. Its probably best for us to use the Bulma version that is specified in the Buefy dependencies so that we are using the version they test against. I haven't run into any incompatibilities yet, but figured this would be safer until there's a reason to pull in the latest Bulma.
2. The css overrides for SimpleForm+Bulma were defined globally and broke Buefy. Since they should only be needed within a SimpleForm, i've scoped them accordingly. Note: i didn't test this because i'm not familiar with the specific tweaks. LMK if this works, @maebeale .

### DIFF
--- a/app/assets/stylesheets/simple_form_customization.scss
+++ b/app/assets/stylesheets/simple_form_customization.scss
@@ -1,14 +1,17 @@
 //# hacks for simple-form + bulma
 
-.is-inline-flex {
-  display: inline-flex;
-}
-.select {
-  vertical-align: baseline;
-}
-.form-actions {
-  padding-top: 3em;
-}
-.select:not(.is-multiple):not(.is-loading)::after, .navbar-link:not(.is-arrowless)::after {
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/simple_form_customization.scss
```diff
@@ -1,14 +1,17 @@
...
-}
-.select {
-  vertical-align: baseline;
-}
-.form-actions {
-  padding-top: 3em;
-}
-.select:not(.is-multiple):not(.is-loading)::after, .navbar-link:not(.is-arrowless)::after {
-  margin-top: 2.5em;
+.simple-form {
```
exbinary, line 3: Only change was to wrap everything here under `.simple-form`. All the other diff lines are just indenting.

PR #80, opened 2020-04-29
Various test cleanups

h-m-m:
https://github.com/maebeale/mutual-aid/pull/75 is big, so I'm breaking things out of it

@exbinary's commit feels uncontroversial to me.

Also, we have an option to import `assert` in setup.js so you don't have to manually import it. I like removing the boilerplate. I'm worried it might confuse new people, and I'm sad that it seems to hamper some auto-complete functionality in Visual Studio Code — but only some. Any thoughts?


spec/factories/listings.rb
```diff
@@ -1,7 +1,6 @@
 FactoryBot.define do
   factory :listing do
     location
-    service_area
```
maebeale, line 4: per discussion, we still need this

spec/factories/listings.rb
```diff
@@ -1,7 +1,6 @@
 FactoryBot.define do
   factory :listing do
     location
```
maebeale, line 3: i'll check out how to make sure listing db/model don't require location

spec/factories/listings.rb
h-m-m: @maebeale If it's OK with you, I want to merge this in first. Then we can add this back to the factory once the db supports it

spec/factories/listings.rb
h-m-m: It's already optional in the db, so that's good. I'll update the factory accordingly

PR #82, opened 2020-05-02
Add more data to tiles

h-m-m:
I've got this so it looks fine on mobile, and on larger screens. There's some weirdness at intermediate sizes, which I'll post screenshots at below. I'm still learning a lot about CSS, haven't really done much with flexbox before now, so suggestions are welcome.

Also, I'm not really happy with how the content of these `<style>` tags in the components look. The feel rambling and hard to scan to me. I'd definitely appreciate organizational suggestions for them

## Problems with intermediate sizes

Not sure how best to deal with this (see pointy arrow) behavior where the tags are too big to fit. I figure it's a UI problem because for some people it will break their ability to scan through the cards easily.

Screen Shot 2020-05-01 at 8 22 17 PM

I like wrapping better than the ol' css overflow.

css-mug-3

I think it's more important to move on to filtering functionality first. I mean, I could keep going with these tiles and fiddle around with various properties to get the behavior I'm looking for. There are also other UI touch-ups I'd like to do. And there are more that we'll notice or stakeholders will ask for. I have a soft preference to do them in a batch later instead of now when I feel like I'm at a good transition point.

## Mobile

Screen Shot 2020-05-01 at 8 19 14 PM

## Large

Screen Shot 2020-05-01 at 8 29 54 PM

### DIFF
--- a/spec/javascript/pages/browse/ContactIcons.spec.js
+++ b/spec/javascript/pages/browse/ContactIcons.spec.js
@@ -3,7 +3,7 @@ import {shallowMount} from '@vue/test-utils'
 import ContactIcons from 'pages/browse/ContactIcons'
 
 it('generally works', function() {
-  var wrapper = shallowMount(ContactIcons, {
+  const wrapper = shallowMount(ContactIcons, {
     propsData: { contactTypes: [{id: 1, name: "email"}, {id: 2, name: "text"},{id: 3, name: "call"}]}
   })
   assert.exists(wrapper.find('b-icon-stub[aria-label=email][icon=envelope]').text())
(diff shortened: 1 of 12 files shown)


maebeale:
Looks AMAZING!!! Agree -- let's ship it.

spec/javascript/pages/browse/ContactIcons.spec.js
```diff
@@ -0,0 +1,11 @@
+import Vue from 'our_vue'
```
h-m-m, line 1: Even though I don't reference it directly, this seems to make the test play nice with Buefy. The test fails if I remove it, even with using `shallowMount`

h-m-m:
I pulled out the biggest remaining TODO into https://github.com/maebeale/mutual-aid/issues/100

PR #85, opened 2020-05-02
Add basic support for docker

grenewode:
The `docker-compose` file is able to launch the rails application and
connect it to a postgres database, also running in docker. However, no other services are
working.

In particular, there is no ability to send emails at this time.
This means that I have not been able to create new user accounts, so much
of the core functionality of the application has not been tested.

This creates a new configuration which loads information from docker
environment variables.

To test this on your machine, you must
1. Have docker-compose installed and
2. You will need to generate a new `SECRET_KEY_BASE`, using the `rake secret` command.
3. The output of that command should be placed in a `.env` file in the
`docker/development/` folder. The `.env` should look like this:
```bash
SECRET_KEY_BASE=e0517d0887e68ebc518600...
```
4. That should be everything you need to get started. Just run
`docker-compose up --build` in the `docker/development/` directory, and
docker will build & launch the project.
5. You should be able to access the project at http://localhost:3000/


grenewode:
I've never worked with Ruby or Ruby on Rails before, so I've got this working somewhat but I'm not sure if I've done it in the best way.

maebeale:
@grenewode hi!!!

maebeale:
yes, i thought i'd mentioned on the call we hadn't yet connected mail services. that's on my weekend to-do list. if you run rake db:seed, you get a system user (test email and password are in `db/seeds.rb`), so that's what we've been for that portion.

maebeale:
we're using Devise (https://github.com/heartcombo/devise) for account management. 

they have a number of strategies for invitations, password recovery, account locks, etc. would love your thoughts on how you think the account functionality should work as we have lots of options available to us. i have some favorites, but would be great to get your perspective.

maebeale:
great job on getting `docker-compose` working! will check this out.

grenewode:
> yes, i thought i'd mentioned on the call we hadn't yet connected mail services. that's on my weekend to-do list. if you run rake db:seed, you get a system user (test email and password are in `db/seeds.rb`), so that's what we've been for that portion.

Thanks for the tip! I'll see if I can get that set up.

maebeale:
@grenewode looks like you'd added `db:setup` in the docker config, so that user should be there without your needing to explicitly run `db:seed` (it's included in `db:setup`)

maebeale:
@grenewode fyi that we've been doing some db restructuring this weekend that includes some breaking changes, so there will likely need to be one db reboot in the next day or so (and then going forward we should be able to do all db changes through the db/migrate DDL files)

maebeale:
LGTM so far, btw, but will circle back and look at this more closely tomorrow

maebeale:
@grenewode would you consider including an update to the readme that references the docker-compose option?

grenewode:
@maebeale I just added some instructions in the README. Let me know if I put them in the right place, or if there's anything I should expand on

grenewode:
I also added a call to `rake db:seed` so the default user is created - and seems to be working! Thanks for your advice getting that setup. I think you probably did mention it in the call, I suspect I just forgot.

maebeale:
Glad it worked! Sure thing! Plenty of sweeping to do, but there's quite a bit stubbed out.

I've got several PRs in the works. Looking forward to your finds and suggestions! Feel free to add issues, tho just know we weren't planning for anyone to have to deal with our current state until after this week's push. :) Given the current state described on the last call though, it seemed worth it to get y'all going on standing the thing up so we can deploy (a better version) asap. 

Our main goals for this week are forms that can accept data (that won't get lost like in googleland) and filterable results. That will get you to a better current state asap. Sounded like everything else is secondary to that. Do you agree?

grenewode:
@maebeale Getting forms working seems like a great thing to focus on - and the rest of the app is actually looking pretty good. I'm working on some stuff to get things working for nixos, and I can make a pull request for that, but I'm not sure that anyone else would use it and it might just complicate the repository

maebeale:
@grenewode cool re nixos. I'm definitely open to checking it out. Curious to see what was needed to pull it off, but, yeah, we can make the call later whether it's a merge vs remains a fork.

h-m-m:
Hi! I'm so happy to see someone else who wants to contribute to this project!

I was able to get this to work for me. The database didn't get seeded right at first, but I think that was because I built the image yesterday based on an earlier version of this PR from before you added the `db:seed` step.

I was also able to run `docker-compose run mutual-aid-web-app rails c` to manually add data without any problem, so that's quite nice

After you're happy with this and are ready to mere it in, do you mind modify the bin/setup script? I'm thinking we can have it accept `bin/setup --with-docker` or something like that and it will go through the some of the setup steps you've documented. The nice thing about that is it will allow us to tweak the docker commands in the script without a need to regularly update the documentation to match

grenewode:
@h-m-m the setup issues could be a result of having the earlier version. You can wipe all the state of the services this sets up with `docker-compose down --volumes --remove-orphans`, although it sounds like you worked it out :). I'm not a docker pro though, so there might be issues. 

Also, the `bin/setup` file looks great! I think it might be able to replace the `entrypoint.sh` script altogether.

I do have a couple of questions
1. What is the difference between `db:prepare` and `db:setup`, `db:seed`? I'm not sure which a docker setup script should call.
2. Being new to ruby on rails, I'm not sure if there's a standard way to parse command line inputs. Is there something in the ruby standard library, or a library that this project is using? I want to keep things as canonical as possible.

I also wanted to say that this project is really a pleasure to work with so far - even though I've never worked with Ruby on Rails or Postgres before, I've been able to get started working on it easier than some C++ and nodejs projects I've encountered.

maebeale:
@grenewode here's the rails code for db rake* tasks: https://github.com/rails/rails/blob/6-0-stable/activerecord/lib/active_record/railties/databases.rake
(*Ruby's "make" is called "rake", tho these days they use the "rails" alias a lot) 

* `db:prepare` - "Runs [db:]setup if database does not exist, or runs migrations [db:migrate] if it does"
* `db:setup` - "Creates the database, loads the schema, and initializes with the seed data (use `db:reset` to also drop the database first)", so, `db:create db:schema:load db:seed`
* `db:seed` - runs the `db/seeds.rb` file

An important difference here is between `db:schema:load` and `db:migrate`:
`db/migrate/*` is where all the DDL "migration" files are stored that create tables, indexes, fks, etc, so the usual pattern is 1. `db:create`, to create the db (will fail if it exists), 2. `db:migrate`, to run all the migrations sequentially, and 3. `db:seed` which runs the seeds.rb file. You can run them all sequentially with one line `rake db:create db:migrate db:seed`

`db:schema:load` builds the database relations, etc, from whatever DDL is declared in the `db/schema.rb` file (which gets updated every time you run `db:migrate`). It doesn't care whatever files are in the `db/migrate` folder. But it drops the db schema before loading, so it'll wipe your data each time.

An interesting note is that projects that last a good while tend to be harder and harder to generate new db's from only the migrations  -- models (class definitions) get deleted and renamed, etc. So generating it from the existing compilation of DDL is the way to go there. Usually there's a `structure.sql` file in the db folder and no `schema.rb` at that point, and `db:structure:load` happens through it rather than schema.rb.

Anyways, `db:prepare` doesn't end up running seeds if the database already exists (I didn't know that!), and db:setup always does, and it doesn't wipe the data in your database the way db:schema:load does. My vote would be to put `db:prepare` in the `docker-compose` bc it allows for the retention of the db. Seeds generally are written so they're idempotent, so it's supposed to be ok if they get re-run, but ideally you want to only create your db the first time and then manually run `db:drop db:create db:migrate db:seed` if you want to clear it out.

Not sure if that's the granularity of answer you were looking for!

maebeale:
For question 2, you know, I'm not sure I know the official answer. I've written custom rake tasks and parsed command line arguments for them, but haven't done it otherwise.

h-m-m:
@grenewode If you want to do processing of potentially complex command line arguments, the most common way to handle this in Ruby is to use a gem called `thor`. Thor is already referenced in this project because both Rails itself and the RSpec test tools rely on it, so you would not have to install anything new to use it.

As someone who's been programming with Ruby on Rails for a while but hasn't done much with thor, I find the documentation at http://whatisthor.com/ a pretty helpful guide to the basics. I can imagine some of it may be a little unusual if you're used to a very different language, so I'm happy to sit down and puzzle anything out with you if you have further questions about it.

grenewode:
@h-m-m Thanks! Looking at other ruby on rails docker projects, it seems that most of them are just using `/bin/sh` as an entry point, and expect you to use `docker-compose run service cmds` to do any manual setup, so for now that's the approach I'm taking. 

In the future we may be able to figure out a nice to way automatically run the `db:prepare`, but for now I'd like to get this merged sooner if possible so that people from LAMA can start testing the application easily.

grenewode:
@maebeale I think I'm feeling pretty good about this right now - if you see other changes that you'd like to make, I think it can be merged. Although it would be good to test it one more time from someone else's computer first, just in case.

maebeale:
Great! Yeah, we agreed this morning we thought it is ready to merge also.
@h-m-m, are you willing to give it one more spin?

grenewode:
I did notice that the app doesn't seem to have a "sign out button"? I'm not sure if I just didn't know were to look though. How do you want things like that reported? Is there a template for feature requests in the issues?

maebeale:
@grenewode there should be a "Log in/out" button in far right of nav ?

maebeale:
Let me check in with the others on preferred issue tracking and will reply w that.

PR #87, opened 2020-05-04
Sendgrid integration

maebeale:


.circleci/config.yml
```diff
@@ -43,6 +43,7 @@ commands:
           command: |
             set -euxo pipefail
             bundle config path .bundle
+            cp .env.ci.example .env.test
```
maebeale, line 46: @exbinary are you comfortable w having changed `.env.example` to `.env.ci.example` ? 
(@h-m-m thought it makes it more obvious that changes here could affect the circleci build)
- .env.example is checked in to source control. 
- we had to perform this copy during circleci build, bc it needed the env vars to complete seeds.rb.

PR #89, opened 2020-05-04
Update communication log fields

maebeale:


app/models/communication_log.rb
```diff
@@ -1,5 +1,9 @@
 class CommunicationLog < ApplicationRecord
   belongs_to :person, optional: true
   belongs_to :match, optional: true
+  belongs_to :created_by, optional: true, class_name: "User", foreign_key: "created_by_id"
 
+  DEFAULT_DELIVERY_STATUS = "completed"
+  AUTO_DELIVERY_CHANNEL = "autoemail"
+  DELIVERY_CHANNELS = [AUTO_DELIVERY_CHANNEL, "email", "text", "call", "snailmail"]
 end
```
maebeale, line 9: Not sure if y'all are ok w this pattern of using model constants -- here or elsewhere in the app. Would love your thoughts.

PR #91, opened 2020-05-04
Send autoemails and log after submission create

maebeale:
This PR creates plumbing for sending confirmation emails after form "Submissions" are created, and then logging the autoemail in a `CommunicationLog`. It includes `SubmissionMailer` mailer file and view templates.

This process of sending autoemails requires some built-in knowledge within the system of who within the `Organization` should get notified of new `submissions` and be suggested to an unauthenticated user  as a point of contact for questions.

It starts to tease out some more detail wrt `SystemSettings`, system `exchange_types`, and the overall workflow, which will later inform a state machine on `Listing`.

* Add mailer and views for confirmation emails
* Change `SubmissionsController` create action so emails are sent after `@submission.save`
* Add `DeliverNowWithErrorHandling` concern to send emails and catch errors
* Add method to `communication_log.rb` to create submission confirmation email log
* Change order on `CommunicationLogsController` and `SubmissionsController` indexes so newest is on top
* Change display of body on `CommunicationLog` and `Submission` `form` and `show` partials/pages
* Change `seeds.rb` to first_or_create a current_organization and a `SystemSetting` record, and to create a Position for each form_type: `ask`, `offer`, `community_resource`
* Add class and instance methods to `Organization` to find who to tell users to contact with questions, and from whom emails will be "sent"
* Allow `Position` to be created even when there isn't a designated `Person` in the role
* Add array of possible `position_type` values (`FORM_CONTACT_TITLES`) -- this could def have a better name
* Added start of `EXCHANGE_TYPE` values to `SystemSetting`, which is also the beginnings of getting a solid state machine workflow in place -- lots of questions here around possible settings to provide to instances. Also added some helper methods for these settings so conditional logic for emails can be implemented.
* Change view_header to only show form if `view_action_name == "edit"`
* Change `edit_button` helper to allow text override (so it can be used on Submissions index to view the related communication_log -- rn that's accessed via person, and that's a little wonky. we might want to associate logs directly to Submission?)

### DIFF
--- a/app/controllers/concerns/deliver_now_with_error_handling.rb
+++ b/app/controllers/concerns/deliver_now_with_error_handling.rb
@@ -1,0 +1,16 @@
+module DeliverNowWithErrorHandling
+  extend ActiveSupport::Concern
+
+  def deliver_now_with_error_handling(autoemail, mailer_view_name=nil)
+    delivery_status = CommunicationLog::DEFAULT_DELIVERY_STATUS
+    begin
+      autoemail.deliver_now
+    rescue => e
+      delivery_status = "FAILED #{e}"
+      Rails.logger.info("=============BEGIN ERROR/PROBLEM WITH EMAIL DELIVERY: #{mailer_view_name}=================")
+      Rails.logger.info("============= #{e} =================")
+      Rails.logger.info("=============END ERROR/PROBLEM WITH EMAIL DELIVERY: #{mailer_view_name}=================")
+    end
+    delivery_status
+  end
+end
--- a/app/mailers/submission_mailer.rb
+++ b/app/mailers/submission_mailer.rb
@@ -6,13 +6,16 @@ class SubmissionMailer < ApplicationMailer
   def new_submission_confirmation_email(submission)
     @submission = submission
 
+    current_organization = Organization.current_organization
+    @organization_name = current_organization.name
+
     @form_name = @submission.form_name
     if @form_name.downcase.include?("ask")
-      @form_contact = Organization.current_organization.ask_form_contact
+      @form_contact = current_organization.ask_form_contact
     elsif @form_name.downcase.include?("offer")
-      @form_contact = Organization.current_organization.offer_form_contact
+      @form_contact = current_organization.offer_form_contact
     elsif @form_name.downcase.include?("community_resource")
-      @form_contact = Organization.current_organization.community_resources_contact
+      @form_contact = current_organization.community_resources_contact
     end
 
     @person = @submission.person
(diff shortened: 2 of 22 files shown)


app/controllers/concerns/deliver_now_with_error_handling.rb
```diff
@@ -0,0 +1,16 @@
+module DeliverNowWithErrorHandling
```
maebeale, line 1: Flag/reminder that we might not want to keep this implemented as a controller concern

app/controllers/concerns/deliver_now_with_error_handling.rb
```diff
@@ -0,0 +1,16 @@
...
+  extend ActiveSupport::Concern
+
+  def deliver_now_with_error_handling(autoemail, mailer_view_name=nil)
+    delivery_status = CommunicationLog::DEFAULT_DELIVERY_STATUS
+    begin
+      autoemail.deliver_now
+    rescue => e
+      delivery_status = "FAILED #{e}"
+      Rails.logger.info("=============BEGIN ERROR/PROBLEM WITH EMAIL DELIVERY: #{mailer_view_name}=================")
```
maebeale, line 10: TODO - review logger patterns/settings/server log retention

app/mailers/submission_mailer.rb
```diff
@@ -0,0 +1,37 @@
...
+    @submission = submission
+
+    @form_name = @submission.form_name
+    if @form_name.downcase.include?("ask")
+      @form_contact = Organization.current_organization.ask_form_contact
+    elsif @form_name.downcase.include?("offer")
+      @form_contact = Organization.current_organization.offer_form_contact
+    elsif @form_name.downcase.include?("community_resource")
+      @form_contact = Organization.current_organization.community_resources_contact
+    end
```
maebeale, line 16: Not sure if we want to keep this implemented using base seeds, and instance and class methods?

app/mailers/submission_mailer.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    @person = @submission.person
+    @locale = @person.preferred_locale || 'en'
```
maebeale, line 19: TODO - use this info to call different templates based on language preference (and using Mobility fallbacks?)

app/mailers/submission_mailer.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    @system_settings = SystemSetting.current_settings
```
maebeale, line 21: This is a class method that uses the first SystemSetting record as a User-admin-able way to change settings. seeds.rb includes creating one default record. This should be revisited to make sure we're all ok w this implementation.

app/models/system_setting.rb
```diff
@@ -1,2 +1,56 @@
 class SystemSetting < ApplicationRecord
+
```
maebeale, line 2: All the stuff around exchange type and comments below need to be thought through, and will also inform the state machine on Listing (which I started, but we need to more clearly define the workflow before I can move forward on that.)

app/models/system_setting.rb
```diff
@@ -1,2 +1,56 @@
...
+  # people are able to message each other, but can request messages be tracked/moderated
+  # people are not able to contact each other, but can request to be connected with specific contributions
+  # people only contact dispatch
+  #
+
+  def self.current_settings
+    self.last
+  end
+
+
```
maebeale, line 35: Methods below will reflect whatever we decide are the system options

app/views/submission_mailer/_generic_confirmation_email.html.erb
```diff
@@ -0,0 +1,31 @@
+<div class="col-md-12">
+  <% person_first_name = @person&.first_name.to_s %>
+
+  <div><b>Hello there, <%= person_first_name %></b>,</div>
+  <br>
+  <div>Thank you for your response to <%= Organization.current_organization %>'s <%= @form_name %>!</div>
+
+  <div>
+    <% if @submission.moderation_requested? || @system_settings.dispatch_moderated? %>
```
maebeale, line 9: TODO - still need to build out what it means to request moderation. We've got it hinted at on the form designs, but haven't yet worked through what this looks like and the downstream effects.

app/views/submission_mailer/_generic_show_for_email.html.erb
```diff
@@ -0,0 +1,24 @@
...
+      <%= "No email on record." %>
+    <% end %>
+    Phone: <%= @person.phone %></p>
+  <p><%= "(" + @person.location&.full_name + ")" if @person.location_id %></p>
+
+  <br>
+
+  <div>
+    <hr>
+    <%#= render partial: "submissions/submission_responses", locals: {email_display: true} %>
```
maebeale, line 21: Once we're creating submissions where we have the form submission info captured, this would be to show the body of what they submitted.

db/seeds.rb
```diff
@@ -9,4 +9,12 @@
   end
 end
 
+host_organization = Organization.where(is_instance_owner: true).first_or_create!(name: "[CHANGEME] Default Mutual Aid Group")
+
+default_system_settings = SystemSetting.create!
```
maebeale, line 14: Here we're creating that default setting.

app/mailers/submission_mailer.rb
maebeale: (Oops. I split the addition of this field into another PR.)

PR #101, opened 2020-05-05
Send autoemails and log after submission create

maebeale:


maebeale:
This PR was pointing to `send-autoemails-and-log-after-submission-create` as a base, so moving this back to master now.

PR #102, opened 2020-05-05
Translations ui

maebeale:
Adds UI for people to be able to change system-level publishing options for locales and rudimentary start at Translations, as well as some small fixes to templates.

If reviewing, the commits might be better bet bc they separate out the CRUD for `SystemLocale` and `MobilityStringTranslation`.

* `SystemLocale` CRUD to store all languages (and settings re `publish_in_dropdowns`, `publish_translations`)
* Initial `MobilityStringTranslation` CRUD to be able to view/create translations through the UI. 
  * This will need some more work, but is a great start for admins being able to get going on storing translation data so by the time we can support translatable `Forms`, `Announcements` and `CommunityResources`, they have the data ready.
  * I did the CRUD on just the `string_translations` table (and not `text_translations`) bc our default is set to `string_translations`, and I'm not clear on why we would ever choose to use the `text_translations` table since Postgres string and text are both `varlena`, afaik. The current `include Mobility` does have some references to it, but those will need to be tightened up.


db/schema.rb
```diff
@@ -229,6 +229,7 @@
     t.bigint "location_id"
     t.bigint "service_area_id"
     t.bigint "user_id"
+    t.string "preferred_locale", default: "en", null: false
```
maebeale, line 232: this shouldn't be here

app/models/mobility_string_translation.rb
```diff
@@ -0,0 +1,16 @@
...
+  validates :locale, presence: true
+  validates :key, presence: true
+  validates :value, presence: true
+
+  validates :translatable_id, uniqueness: { scope: [:translatable_type, :locale, :key] }
+
+  scope :translatable, ->(record) { where(translatable_id: record.id).where(translatable_type: record.class.to_s) }
+  scope :locale, ->(locale) { where("locale ILIKE ?", locale) }
+  scope :key, ->(key) { where("key ILIKE ?", key) }
+  scope :value, ->(value) { where("value ILIKE ?", value) }
```
maebeale, line 15: Have some starts on param filtering setup for indexes, so this is for sandbox use rn.

app/models/mobility_string_translation.rb
```diff
@@ -0,0 +1,16 @@
+class MobilityStringTranslation < ApplicationRecord
+  
+  belongs_to :translatable, polymorphic: true
+  belongs_to :created_by, class_name: "User", optional: true
+
+  validates :locale, presence: true
+  validates :key, presence: true
+  validates :value, presence: true
+
+  validates :translatable_id, uniqueness: { scope: [:translatable_type, :locale, :key] }
```
maebeale, line 10: This enforces the index constraint set by mobility gem install on the Rails side

db/migrate/20200505095309_update_translations.rb
```diff
@@ -0,0 +1,9 @@
+class UpdateTranslations < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :mobility_text_translations, :created_by, null: true, foreign_table_name: :users, index: true
+    add_column :mobility_text_translations, :is_approved, :boolean, null: false, default: true
+
+    add_reference :mobility_string_translations, :created_by, null: true, foreign_table_name: :users, index: true
+    add_column :mobility_string_translations, :is_approved, :boolean, null: false, default: true
```
maebeale, line 7: These add audit/review options for admins, tho it won't be fully implemented until we have Forms dynamically calling translated content. 

Adding them to both tables for now, bc I'd started out writing a materialized view to combine these, but then backed out and went with just adding a model for MobilityStringTranslations

exbinary:
A couple of things i'd like to discuss:
* I had envisioned content (mobility) translations to happen in-situ on Listing and Comm. Resource pages; so as a user is filling out a form, we'd capture their submissions in their chosen locale, and then provide options to add translations of that content. In my mind, all of that would happen on the /edit page of the corresponding content, not on a separate admin page.
* Something about having `SystemLocales` in both the yaml _and_ in the db seems off?

maebeale:
I think we're saying same thing? Translations table is where the admins create the translated content to display to the end user. And then their responses are stored in their native language. 

Did you mean translations of what people submit?

maebeale:
The SystemLocales table is there so admins can add/remove/edit/publish/unpublish the locales the want to support and related translated content. 

The yml file is to seed the SystemLocales table when we set up the instance so they don't have to manually add every language. But since they'll be changing it, the yml shouldn't get used again. I could add a block around that section to only seed if the table is empty.

Does that sound/feel any better/different?

exbinary:
> I think we're talking about two different things? Translations table is where the admins create the translated content to display to the end user. And then their responses are stored in their native language.

Maybe...? Using the language used by the Mobility author [here](https://dejimata.com/2017/3/3/translating-with-mobility), i see two kinds of translations:
1. _Interface_  -- site nav, form labels, that kind of thing
2. _User submitted content_ -- Offers, Asks, Community Resources, Announcements, maybe Comm. Logs.

Is there a third category?

I've been assuming Mobility would be for user-submitted content and we'd stick with yaml based i18n for interface translations?

maebeale:
Yup. 

I agree with you that we could set up the system to translate user-submitted content on the fly, so no first-glance need for admin UI there. However, they'll need to edit user-generated content, so we do need it for that (Tho that could live on Edit page).

But what I've seen from the groups is they are provided with information in English and they are manually translating that info into other languages after the fact. 

CustomFormQuestions, at minimum, need an admin side foot those translations.

exbinary:
> However, they'll need to edit user-generated content, so we do need it for that (Tho that could live on Edit page).

Yeah, having the translations be on the edit page makes more sense to me because it would give the translator more context. Context is often what makes or breaks translation.

> But what I've seen from the groups is they are provided with information in English and they are manually translating that info into other languages after the fact.

That sounds right. What could work is to have an additional filter on /contributions that admins can use to show which contributions need translations. And the actual translating happens on the /edit page of the content itself.

> CustomFormQuestions, at minimum, need an admin side for those translations.

With these as well, it makes more sense in my brain to have translation be in-situ.

exbinary:
> The yml file is to seed the SystemLocales table when we set up the instance so they don't have to manually add every language. But since they'll be changing it, the yml shouldn't get used again. I could add a block around that section to only seed if the table is empty.
> 
> Does that sound/feel any better/different?

Yes, that makes sense to me now. Thanks for clarifying! A guard around the seed would be good. IIUC, all seeds are supposed to be idempotent, yeah?

maebeale:
```With these as well, it makes more sense in my brain to have translation be in-situ.```

Yeah, I'm coming around to this -- implementing on the Edit pages. I'd still love to retain access to the full table for admins tho. As an admin user I get really antsy when I can't access everything in bulk. Are you ok w keeping it on admin page, but driving most use to the respective Edit pages?

**Comment by maebeale** [comment]
p.s. @exbinary if you still would prefer we not have the Translations admin, I'll def remove it. :)

**Comment by maebeale** [comment]
@exbinary ok to commit this and the admin pages PRs and I'll add a new one w edit page ui?

**Comment by exbinary** [comment]
@maebeale, i think we're aligned now on the things we discussed.

The thing i'm still hesitant about is introducing our own `MobilityStringTranslation` model around the mobility table. Since mobility will manage normal translation operations, i'm not clear yet what additional need there is for the model?

I'm fine with an index view of all translations. Could we get rid of the other actions and limit the route to just index. Each item could like to the content itself for editing.

**Comment by maebeale** [comment]
The model provides us with the ability to use scopes and URL params to search the (and display) index view. 

Will remove the other actions.

**Comment by maebeale** [comment]
Will remove the other actions

**Comment by exbinary** [comment]
> The model provides us with the ability to use scopes and URL params to search the (and display) index view.

I see. I think that may end up making sense, but i'm still hesitant, TBH :).

Mobility ships with its own models that represent the key/value tables:
`Mobility::ActiveRecord::StringTranslation` and `...::TextTranslation`, which define their own validations and polymorphic association. See deets [here](https://github.com/shioyama/mobility/wiki/KeyValue-Backend)

I think it could be confusing to have our own model that duplicates that definition.
I'd vote to start by using the mobility models directly for querying and evaluating how much duplication of querying we see. If its just a few `where` clauses in one or two controllers/interactors, that could be fine. If more, we could consider extending their models for scope definitions.

**Comment by maebeale** [comment]
Agree with all that you're saying. Def re having two "models," but I haven't been able to access those classes or assn methods, nor to figure out how to, on the Edit page, call something along the lines of @record.mobility_string_translations (without a model) to be able to render any existing translations.

And, for adding new ones, unless it happens through a separate controller (to which we could pass the record_class and record_id so the data is shown contextually, like I had in the form), I'm unclear how to do it with accepts nested attributes unless there's a model and assn method and controller to which I can add strong params, etc.

Would LOVE help bc maybe I'm def missing how to do it all without a "normal" rails model.

Screen Shot 2020-05-07 at 2 29 49 PM
Screen Shot 2020-05-07 at 2 37 12 PM

**Comment by maebeale** [comment]
I started to do the changes in this branch, but am instead going to merge this (and another branch which depends on it) for testing by LAMA and will start a new branch re mobility-cleanup.

**Comment by exbinary** [comment]
Oh, i think those models must be lazy loaded somehow:
```ruby
$ mutual-aid: bin/rails c
# Running via Spring preloader in process 85840
# Loading development environment (Rails 6.0.2.2)
irb(main):001:0> Mobility::ActiveRecord::TextTranslation
Traceback (most recent call last):
        1: from (irb):1
NameError (uninitialized constant Mobility::ActiveRecord::TextTranslation)

irb(main):002:0> Listing.count
   (22.2ms)  SELECT COUNT(*) FROM "listings"
=> 1

irb(main):003:0> Mobility::ActiveRecord::TextTranslation
=> Mobility::ActiveRecord::TextTranslation(id: integer, locale: string, key: string, value: text, translatable_type: string, translatable_id: integer, created_at: datetime, updated_at: datetime)
```

PR #104, opened 2020-05-06
Some improvements to our JS specs

exbinary:
1. 0e8085f : Eliminate Vue warning about duplicate keys when running specs
1. a039955 : Ensure all specs have a top-level describe block to help with output grouping
1. 4352e79: Rework vue configuration so it can be safely reused in specs without polluting a shared global Vue instance between specs.

Note: the diffs from number #2 above are mostly just whitespace, so i recommend ignoring whitespace changes on review.

### DIFF
--- a/spec/javascript/pages/browse/ContactIcons.spec.js
+++ b/spec/javascript/pages/browse/ContactIcons.spec.js
@@ -1,10 +1,11 @@
-import Vue from 'our_vue'
-import {shallowMount} from '@vue/test-utils'
+import {configure} from 'vue_config'
+import {createLocalVue, shallowMount} from '@vue/test-utils'
 import ContactIcons from 'pages/browse/ContactIcons'
 
 describe('ContactIcons', () => {
   it('generally works', function() {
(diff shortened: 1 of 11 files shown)


spec/javascript/pages/browse/ContactIcons.spec.js
```diff
@@ -1,11 +1,14 @@
...
+import {createLocalVue, shallowMount} from '@vue/test-utils'
 import ContactIcons from 'pages/browse/ContactIcons'
 
-it('generally works', function() {
-  const wrapper = shallowMount(ContactIcons, {
-    propsData: { contactTypes: [{id: 1, name: "email"}, {id: 2, name: "text"},{id: 3, name: "call"}]}
+describe('ContactIcons', () => {
+  it('generally works', function() {
+    const wrapper = shallowMount(ContactIcons, {
+      localVue: configure(createLocalVue()),
```
exbinary, line 8: `vue-test-utils` pattern for ensuring we don't pollute a shared global Vue instance between specs.

PR #106, opened 2020-05-07
First draft of using interactors for nested forms

exbinary:
Demonstrates using [ActiveInteraction](https://github.com/AaronLasseigne/active_interaction#rails) gem to help manage nested models involved in creating a listing: `Listing -> Person -> Location`.

It may be easier to review each of the two substantive commits separately, because the last one introduces a further abstraction on top of the gem.

While `compose` helps with calling one interactor from another, it doesn't handle two things that are important when dealing with nested forms:
1. merging errors from the child object into the parent
2. rolling back child records if parent records later fail
I've introduced the `merging_errors` method to remove the duplication involved with handling those considerations.

Note: i haven't defined all the necessary fields for any of these models. Just the minimum required to get it working and also test validation/errors.

There's more we could do with the gem. In particular, it supports ActiveModel validations at the interaction layer. I do prefer validations in that layer, but i'm hesitant to do this across the board because many controllers aren't complex enough to need an Interactor.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,50 +3,37 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby '2.7.0'
 
-# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 6.0.2', '>= 6.0.2.2'
 
 # This should after rails but before any gems that might require it
 gem 'dotenv-rails', require: 'dotenv/rails-now'
 
-# Use postgresql as the database for Active Record
-gem 'pg', '>= 0.18', '< 2.0'
-# Use Puma as the app server
-gem 'puma', '~> 4.1'
-# Use SCSS for stylesheets
-gem 'sass-rails', '>= 6'
-# Transpile app-like JavaScript. Read more: https://github.com/rails/webpacker
-gem 'webpacker', '~> 4.0'
 # Turbolinks makes navigating your web application faster. Read more: https://github.com/turbolinks/turbolinks
 # gem 'turbolinks', '~> 5'
-# Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
-gem 'jbuilder', '~> 2.7'
-
 # Use Redis adapter to run Action Cable in production
 # gem 'redis', '~> 4.0'
 # Use Active Model has_secure_password
 # gem 'bcrypt', '~> 3.1.7'
-
-gem 'devise'
-gem 'reform-rails'
-gem 'simple_form'
-
 # Use Active Storage variant
 # gem 'image_processing', '~> 1.2'
+# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
+# gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
-# Reduces boot times through caching; required in config/boot.rb
-gem 'bootsnap', '>= 1.4.2', require: false
-
+gem 'active_interaction'
 gem 'acts-as-taggable-array-on'
-
-# support translations on more than one model
+gem 'bootsnap', '>= 1.4.2', require: false
+gem 'devise'
+gem 'jbuilder', '~> 2.7'
 gem 'mobility', '~> 0.8.9'
-
-# email and sms
+gem 'pg', '>= 0.18', '< 2.0'
+gem 'puma', '~> 4.1'
+gem 'reform-rails'
+gem 'sass-rails', '>= 6'
 gem 'sendgrid-ruby'
+gem 'simple_form'
+gem 'webpacker', '~> 4.0'
 
 group :development, :test do
-  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
   gem 'binding_of_caller'
   gem 'factory_bot_rails'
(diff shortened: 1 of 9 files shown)


maebeale:
@exbinary are we using difft rake versions? looks like this fail is telling us to run bundle install

app/interactors/base_interactor.rb
```diff
@@ -0,0 +1,20 @@
+class BaseInteractor < ActiveInteraction::Base
+  def merging_errors(in_transaction: false)
```
maebeale, line 2: why is default transaction set to false?

app/javascript/entry_points/orientation.js
```diff
@@ -1,4 +1,4 @@
-import Vue from 'our_vue'
```
maebeale, line 1: this is stuff from the other PR, right?

exbinary: Oops, yes sorry. Forgot to change the base. Fixed now.

app/interactors/base_interactor.rb
exbinary: Great question! I guess i was being conservative with resources, but that's not a good reason at all. I'm down to switch it!

exbinary:
> @exbinary are we using difft rake versions? looks like this fail is telling us to run bundle install

Looking into this now. I did mess with `Gemfile` a bit, but mostly just sorting things...

Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby '2.7.0'
-
+gem 'rake'
```
exbinary, line 5: @h-m-m , i think this is a reasonable fix for [this CircleCI failure](https://app.circleci.com/pipelines/github/maebeale/mutual-aid/189/workflows/9b061ef6-7065-46c5-bbc6-100ac979764c/jobs/719), but let me know if you have any concerns about it.

h-m-m: The problem was that the checked-in Gemfile.lock didn't match the Gemfile.lock after doing `bundle install`. I think I can adjust the CircleCI config that makes troubleshooting this easier. Can you push up one of the commits that failed to its own branch so I can test the improved config works?

h-m-m: This change should not be needed. If you revert it and then cherry-pick https://github.com/maebeale/mutual-aid/commit/1188a0e177347caf077859344e38bbd17febe3eb you should get a much more useful error message if the bundling goes wrong

spec/interactors/save_person_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    it 'allows for existing fields to be nullified' do
+      expect(person.location.city).to be_empty
+    end
+
+    context 'when params include a person.id but no location.id' do
+      let(:current_location) { nil }
+
+      it 'fails' do
+        expect { interaction }.to raise_error(/Missing location.id/)
```
h-m-m, line 35: I really like all the test cases you've outlined here. These answered all the questions I had looking at the rest of the code

spec/interactors/save_listing_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+          city: 'Lansing',
+          state: 'MI',
+        },
+      },
+    }}
+
+    it { is_expected.to be_invalid }
+
+    it 'does not create any records' do
+      expect { interaction }
```
h-m-m, line 24: I feel like I only vaguely understand why testing the success case would be unnecessary here. Like, I think it's covered by everything else, but I'm having a little trouble following that line of thought today

app/interactors/save_listing.rb
```diff
@@ -0,0 +1,15 @@
...
+  record :service_area
+  string :type
+
+  hash :person, strip: false
+
+  # todo: add other fields here and in nested interactors
+
+  def execute
+    merging_errors(in_transaction: true) do
+      person_record = compose SavePerson, person
```
h-m-m, line 11: I was worried about the potential for weird nested transaction behavior here until I saw in the ActiveInteraction docs where `compose` will raise an error on failure like `.run!` 👍

spec/interactors/save_listing_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'rails_helper'
+
+RSpec.describe SaveListing do
+  let(:service_area)    { create :service_area }
+  subject(:interaction) { SaveListing.run params }
```
h-m-m, line 5: Very much looking forward to seeing this syntax in use

maebeale:
Yup, saw the alphabetization. :)

spec/interactors/save_listing_spec.rb
exbinary: I think you're right that its worth adding a happy path example here ✅.
Will add on the next pass through.

app/interactors/save_person.rb
```diff
@@ -0,0 +1,28 @@
+class SavePerson < BaseInteractor
+  integer :id, default: nil
+  string :preferred_contact_method  # todo: string?
```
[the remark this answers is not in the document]
exbinary: [This example](https://github.com/maebeale/mutual-aid/pull/106/files/271fdcfc88b06ca0909620dbb9692746c532043c#diff-16f1073379df86cff8f30397d3d38296R72) covers that case, though not explicitly or directly. I could make it clearer by setting the `preferred_contact_method` right in the example instead of relying on it from the earlier defined `params`.

spec/interactors/save_listing_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  describe 'validation' do
+    let(:params) {{
+      type: '',  # only Listing field with validation
```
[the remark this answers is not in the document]
exbinary: Do you mean where `person: {}` or is missing from params completely?

The interactor will reject that pretty early since there's no default configured for the person param; it won't even get to the `execute` method.

I tend to eschew specs that test the documented behavior of third-party libraries, but since we're new to `active_interaction`, there might be value in adding a spec demonstrating that behavior?

Gemfile
exbinary: Thanks for looking into this @h-m-m !
As you predicted, once i corrected the lock file, it works [fine](https://app.circleci.com/pipelines/github/maebeale/mutual-aid/223/workflows/357e2fb2-b5b0-4a1b-a1b9-d1032c60c9f4) w/o needing `rake` to be explicitly defined.
I've dropped that commit.

spec/interactors/save_listing_spec.rb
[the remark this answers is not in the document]
maebeale: Yeah, seems easy to leave off that param when dealing w raw json, but maybe what active_interaction provides is more integration w active_model ?

app/interactors/save_person.rb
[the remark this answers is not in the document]
maebeale: i think it's ok. i'm still not used to reading tests, so pretty sure i just missed that. thanks!

PR #115, opened 2020-05-08
[ON HOLD] Some randoms

maebeale:
Some random changes

- DRY admin pages, update admin links (w params yet to be implemented)
- set_system_setting in ApplicationController

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,11 +3,16 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery with: :exception
   before_action :authenticate_user!
-  before_action :admin_status
+  before_action :set_admin_status
+  before_action :set_system_setting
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -3,13 +3,18 @@ class ApplicationController < ActionController::Base
...
+  before_action :set_system_setting
   around_action :switch_locale
 
-  def admin_status
+  def set_admin_status
     @admin_status = params[:admin] ? YAML.load(params[:admin]) : current_user&.admin? # allows admin user to simulate with param=false
   end
 
+  def set_system_setting
+    @system_setting = SystemSetting.first || SystemSetting.new # should only be one of these records per instance
```
maebeale, line 15: Not sure if this is too global? Like, Maybe we don't want this in ApplicationController ?

PR #116, opened 2020-05-08
[REVIEW] Add .circleci/config.yml

maebeale:
Pretty sure this work has already been merged, but creating a PR to check


maebeale:
@h-m-m It looks like there are a couple extra config lines in this outdated file we may want to incorporate into our latest version of circlci config?

maebeale:
these changes are outdated

PR #118, opened 2020-05-08
Offer form skeleton

exbinary:
Putting up enough of the `/offers/new` form to show how the following pieces fit together:
* `offers_controller`
* `interactors/save_listing` and nested interactors
* `blueprints/listing_blueprint` and nested blueprints (serializers)
* `views/offers/new`
* `entry_points/offer.js`
* `pages/Offer.vue` and child components

The commits here show a full round trip including showing server side errors on the front end.
Would love feedback (now, after this push, or both!), particularly on architectural/design aspects.

Its not able to save successfully yet because the form is missing necessary fields.
I'm working on adding the remaining fields now.

Note: i've kept this work separate from `listings_controller` and related views until its ready to be merged with it. Also dealing with only offers first but could become more generic as we refactor to add asks.


maebeale:
When I go to /offers/new, i don't see anything ?

exbinary:
Promoting this from draft to ready for review.

I've gotten it working with `acts_as_taggable_on`, pulling the list of available categories and saving them as tags.

Full round trip works on most fields (`preferred_contact_method` isn't retaining when there are errors). And submission saves a Listing, Person and Location.

Since we have the browser doing required checks, its not straightforward to test errors.
The easiest way is to comment out [this line](https://github.com/maebeale/mutual-aid/pull/118/files#diff-9f0f15d857a397cda45bf0b34f52f787R8) to remove the state field which is currently required on the backend.

app/models/ask.rb
```diff
@@ -1,5 +1,4 @@
 class Ask < Listing
-  belongs_to :receiver, class_name: "Person", foreign_key: "person_id"
```
exbinary, line 2: @maebeale , i don't remember what, but this was causing a problem.

app/models/offer.rb
```diff
@@ -1,5 +1,4 @@
 class Offer < Listing
-  belongs_to :provider, class_name: "Person", foreign_key: "person_id"
```
exbinary, line 2: Same as above.

maebeale: k

app/models/ask.rb
maebeale: k

maebeale:
AMAZING, @exbinary !

PR #119, opened 2020-05-08
Improve circleci bundle install error messages

h-m-m:
This helps with troubleshooting Circle CI problems when the Gemfile and Gemfile.lock don't mach. Before this change, you would only see a failed step once the `ruby_tests` job started to run and the error for that failed step would me misleading as to the cause of the problem.

With this change, the error will show up in the `bundle_install` step, and will be more explicit about what the discrepancy is.

Here is a screenshot of what that more helpful error will look like:

Screen Shot 2020-05-07 at 4 01 08 PM


maebeale:
AWESOME!

PR #126, opened 2020-05-09
ContactMethod model

exbinary:
Refactors 'contact methods' into its own table-backed model.
Motivations:
1. I was finding that i needed additional data for each contact method, such as the name of the corresponding field, and label (though i ended up getting that from `i18n` instead).
1. I could see an settings page where a system admin enables/disables the available contact methods and adds their own ones.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -10,6 +10,7 @@ class PeopleController < ApplicationController
 
   def new
     @person = Person.new
+    set_form_dropdowns
   end
 
   def edit
@@ -59,7 +60,7 @@ class PeopleController < ApplicationController
 
     def set_form_dropdowns
       @system_locales = SystemLocale.where(publish_in_dropdowns: true).pluck(:locale_name, :id)
-      @preferred_contact_methods = Person::PREFERRED_CONTACT_METHODS
+      @preferred_contact_methods = ContactMethod.enabled
     end
 
     def person_params
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -2,6 +2,7 @@ class Person < ApplicationRecord
   belongs_to :user, optional: true, inverse_of: :person
   belongs_to :service_area, optional: true
   belongs_to :location, optional: true
+  belongs_to :preferred_contact_method, class_name: 'ContactMethod'
 
   has_many :communication_logs
   has_many :donations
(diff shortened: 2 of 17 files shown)


spec/interactors/save_listing_spec.rb
```diff
@@ -2,14 +2,15 @@
...
+  let(:contact_method)  { create :contact_method_email }
   subject(:interaction) { SaveListing.run params }
 
   describe 'validation' do
     let(:params) {{
       type: '',  # only Listing field with validation
       service_area: service_area.id,
       person: {
-        preferred_contact_method: 'email',
+        preferred_contact_method: contact_method.id,
```
maebeale, line 13: why is this `contact_method.id` here, and not `contact_method`?

app/views/people/_form.html.erb
```diff
@@ -9,7 +9,7 @@
     </div>
 
     <div class="field is-grouped">
-      <%= f.input :preferred_contact_method, as: :select, collection: @preferred_contact_methods %>
+      <%= f.input :preferred_contact_method, as: :select, collection: @preferred_contact_methods, label_method: :name, value_method: :id %>
```
maebeale, line 12: totally!

app/models/person.rb
```diff
@@ -35,10 +32,9 @@ def temporary_offer_tag_list # these will be directly on the person once forms a
...
 
-  private def validate_preferred_contact_method_data
-    preferred_contact_data = PHONE_CONTACT_METHODS.include?(preferred_contact_method) ? "phone" : preferred_contact_method
-    if preferred_contact_method.present? && !self.public_send(preferred_contact_data).present?
-      errors.add(:person, " -- #{preferred_contact_method} needed (preferred contact method)")
-    end
+  private def preferred_contact_method_present!
+    return unless preferred_contact_method
+    field = preferred_contact_method.field
+    errors.add(field, :blank) if self[field].blank?
```
maebeale, line 38: YESSSS! SO much improved!!!

app/controllers/people_controller.rb
```diff
@@ -10,6 +10,7 @@ def show
 
   def new
     @person = Person.new
+    set_form_dropdowns
```
maebeale, line 13: You rock!

maebeale:
How come the reverse model assns wasn't included in this? (ContactMethod `has_many :people`)
Do you wait to add those until called? I can see us wanting a stat on how many people prefer what...

maebeale:
If we don't know someone's preference, should we allow an "unknown" option that doesn't validate a value is present? (Groups will need to import data, but maybe we can assume it will all have phone or email? And, if we're listing someone as having a position, maybe we don't know their contact info yet?)

Just imagining being an admin and typing 12345 in phone bc it was required.

spec/interactors/save_listing_spec.rb
exbinary: The interactor allows for either an id or actual record (because its defined as a `record` type there), but the POST params will have the `id` so wanted to make that part of the test.

app/models/person.rb
exbinary: 🙌🏾

exbinary:
> How come the reverse model assns wasn't included in this? (ContactMethod `has_many :people`)
> Do you wait to add those until called? I can see us wanting a stat on how many people prefer what...

I do tend to wait on them until they're needed (counts, eg are so easy to do w/o the association on the model), but don't have a strong preference about this.

spec/interactors/save_listing_spec.rb
maebeale: ah! awesome.

PR #127, opened 2020-05-09
Extract AuthTokenInput for reuse on any form

exbinary:
Tiny PR extracting an `AuthTokenInput` component.

### DIFF
--- a/app/javascript/components/NavBar.vue
+++ b/app/javascript/components/NavBar.vue
@@ -30,7 +30,7 @@
 
 <script>
 import logo from 'images/logo.png'
-import DeleteButton from 'components/DeleteButton'
+import DeleteButton from './DeleteButton'
 
 export default {
   props: {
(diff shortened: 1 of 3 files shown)


app/javascript/components/NavBar.vue
```diff
@@ -30,7 +30,7 @@
 
 <script>
 import logo from 'images/logo.png'
-import DeleteButton from 'components/DeleteButton'
+import DeleteButton from './DeleteButton'
```
maebeale, line 33: is this from another PR?

maebeale, line 33: Just curious -- why is this needed/available now?

exbinary: It was not part of the other changes here. I just noticed that `NavBar` and `DeleteButton` are siblings, so this is simpler and helps IDEs 'go-to' the file (they often can't figure out where `components/` is).

PR #130, opened 2020-05-09
Basic filtering UI

h-m-m:
### Description

Add basic filtering functionality

### Follow-up issues ideas prompted by this PR

- UI: have the category and urgency tags always be two rows on tablet to solve the problems with intermittent wrapping due to variable tag list length
- UI: Improve the visual indicators on the filter category expand/collapse accordion
- UI/functionality: allow a category checkbox that toggles all of the subcategories (to select all urgency levels simultaneously, for example)
- Add categories that should be checked by default (can be fed from the Rails side)

#### Mobile breakpoint
Screen Shot 2020-05-09 at 11 31 33 AM

#### Tablet breakpoint

Screen Shot 2020-05-09 at 11 32 30 AM

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,6 +16,7 @@
         "--webpack-config", "config/webpack/test.js",
         "--require", "spec/javascript/setup.js",
         "--timeout", "999999",
+        "-u", "bdd-lazy-var/global",
         "--colors",
         "${file}"
       ],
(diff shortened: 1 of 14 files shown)


maebeale:
Should we be seeing this in the server log? 
Screen Shot 2020-05-10 at 1 10 05 AM

maebeale:
And, I think my fake data might be bad?

Screen Shot 2020-05-10 at 1 11 38 AM

h-m-m:
> Should we be seeing this in the server log?
> Screen Shot 2020-05-10 at 1 10 05 AM

I'll definitely look into this

h-m-m:
> And, I think my fake data might be bad?
> 
> Screen Shot 2020-05-10 at 1 11 38 AM

This looks like you're on the `rails-filtering-api` branch, which I'm not finished with quite yet, instead of the branch in this PR

PR #145, opened 2020-05-10
Change communication_logs.body -> string (and read-only if autoemail)

maebeale:


db/schema.rb
```diff
@@ -156,6 +156,9 @@
     t.string "neighborhood"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.string "locationable_type", null: false
+    t.bigint "locationable_id", null: false
+    t.index ["locationable_type", "locationable_id"], name: "index_locations_on_locationable_type_and_locationable_id"
```
h-m-m, line 161: I think this schema change is from another PR and is why the tests are failing. I wonder if this is also the case for your other PR that's failing mysteriously.

While you're making a lot of schema changes, I wonder if it's faster to not check in this schema file, let CircleCi do the migrations, and then check in the schema.rb once things are more likely to be stable

maebeale: Ooo! Thank you! Yeah, I was wondering how you best handle. Still working through how to get my circumambulation into distinct PR's. Appreciate you SO much, @h-m-m !

PR #151, opened 2020-05-10
Adds `acts_as_taggable_on`

exbinary:
Switches from the array column based `acts_as_taggable_array_on` gem to the join-table based `acts_as_taggable_on`.

When we discussed this some days ago, we'd floated the idea of configuring the gem to point at our `categories` table. I got that to work now that i better understand how the library works, i'm leaning towards sticking with their default `tags` and `taggings` tables.

We'd still keep the `categories` table. That would be the authoritative source of 'available' tags, configurable by system admins. The tag tables would be used exclusively by the gem to track which categories get tagged.

I think that sets us up to support user-provided tags that don't pollute the admin-managed categories table, and avoids any potential collisions between our own migrations and the gem's.

LMK your thoughts on that @maebeale .

I tried to refactor anything that looked to be using `tags` but i'm not confident that i knew what to test 🙈 .


maebeale:
YAYYYY!!!!! Yup, this was my original imagined structure, so all green lights from me! :)

maebeale:
No worries on testing! Yeah, some of it I've just hard-coded until we had a solid strategy for storing and accessing them.

config/locales/en.yml
```diff
@@ -34,3 +34,8 @@ en:
     label:
       listing:
         tags: "What types of support can you offer?"
+  models:
+    contact_method:
+      labels:
+        phone: Phone number
+        email: Email address
```
maebeale, line 41: I changed this to attributes -- is that ok, or did you def want this to remain "labels" ?

PR #160, opened 2020-05-10
Rails filtering api

h-m-m:
### Description

With the addition of the Rails API side of things, it now works round-trip

filter-usage-lowres

Note: the timestamps in the image are `created_at` and not a `publish_until`, as that's not implemented yet

### Implementation details

Any class that responds to `as_filter_types` to return a list of objects that also respond to `name` and `id` can then be used to populate the filter list. Let's call this filter class "FilterClassA". It's then expected that all of the contribution types know what to do when passed `.filter_by(filter_class_a: [1,2,3])`. I'm not too happy with how the `filter_by` side looks, but I think it's fine for now. 

I am happy with the Blueprinter stuff. Yay!

The Vue side has room for multiple contact types, but we're allowing one to be highlighted on the Rails side. This may be fine, actually, and require no immediatee change.

### Reasonable next steps

- Implement the data commented out as "not yet implemented" in contribution_blueprint_spec.rb (urgency, availability, publish_until)
- Improve the experience of what boxes are checked by default. Currently that's none of them, which works the same as checking all if and only if none of the boxes were checked when you first loaded the page, which is definitely wonky and needs to be changed
- Allow filters changes to update the URL so that copying the current page address or reloading the page doesn't lose the filters you changed
- Make the buttons at the bottom of each contribution do something

### Smaller potential follow-up tasks

- Cleaning up some of the tag code to be more readable as we switch to acts_as_taggable
- Add additional filter types
- Complete the list of preferred contact type icons
- UI tweaks mentioned in #130 
- Transition smoothing? Better UI when there's only one or two Tiles?
- Error handling when updating the filter fails (and maybe reverting the checkboxes back so the user can try the same request again?)
- Date format parsing on js side (might be better to pass seconds since epoch from Rails and then use [`setTime()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setTime))
- Add more tests, or break up existing tests

Please comment if you can think of some good tests to add here


app/javascript/pages/Browse.vue
```diff
@@ -4,7 +4,7 @@
       <section class="column is-one-quarter">
         <BrowserSelector :browser="browser" @clicked="browser = $event" />
 
-        <Filters :filterCategories="filterCategories" v-model="activeFilters" />
+        <Filters :filterTypes="filterTypes" v-model="activeFilters" />
```
maebeale, line 7: THANKS again for renaming these, @h-m-m !

app/javascript/pages/browse/TileListItem.vue
```diff
@@ -45,17 +47,17 @@ export default {
     service_area: {type: Object, default: null},
     short_title: String,
     description: String,
-    publish_until: String,
```
maebeale, line 48: Why did we remove publish_until?

app/models/category.rb
```diff
@@ -27,6 +27,10 @@ class Category < ApplicationRecord
       ['cash', ''],
   ]
 
+  # Currently only filtering by top-level categories
```
maebeale, line 30: 👍

app/models/listing.rb
```diff
@@ -17,6 +17,22 @@ class Listing < ApplicationRecord
   enum state: { received: 0, fulfilled: 1 }
 
   def name
-    "#{type}: #{tags}"
+    "#{type}: #{tags.join(', ')}"
```
maebeale, line 20: I think this needs to be `all_tags_list.join`
I recently added some methods like this also, tho not duplicative. Still, prob should review next week (i'll add an issue).

app/models/listing.rb
```diff
@@ -17,6 +17,22 @@ class Listing < ApplicationRecord
...
+  end
+
+  def categories_for_tags
+    Category.where(name: tags)
+  end
+
+  def self.filter_by(conditions, default_scope = nil)
+    scope = default_scope
```
maebeale, line 28: Maybe we

app/models/service_area.rb
```diff
@@ -14,6 +14,8 @@ class ServiceArea < ApplicationRecord
 
   validates :name, presence: true
 
+  scope :as_filter_types, -> { i18n.select :id, :name }
```
maebeale, line 17: oo! cool!

app/serializers/contribution_blueprint.rb
```diff
@@ -0,0 +1,13 @@
+class ContributionBlueprint < Blueprinter::Base
```
maebeale, line 1: YESS!

spec/javascript/pages/browse/ContactIcons.spec.js
```diff
@@ -6,9 +6,9 @@ describe('ContactIcons', () => {
   it('generally works', function() {
     const wrapper = shallowMount(ContactIcons, {
       localVue: configure(createLocalVue()),
-      propsData: { contactTypes: [{id: 1, name: "email"}, {id: 2, name: "text"},{id: 3, name: "call"}]}
```
maebeale, line 9: why didn't you want to use `call` anymore? 
my earlier thinking (open to being convinced otherwise!) was that if ppl say phone, they could maybe think it means text ? (harder to do tho, when the list is right there...)

maebeale:
I was able to get this branch rebased on all the latest, except for one chicken egg situation w ServiceArea requiring a Location. It works on the admin side, but was tough to figure out how to get it working on the Vue/public (and tests) side. Got close! 3 tests failing -- all to do w this.

app/models/service_area.rb
```diff
@@ -3,6 +3,8 @@ class ServiceArea < ApplicationRecord
   translates :name
   translates :description, type: :text
 
+  after_initialize :create_location_if_not_exists
```
maebeale, line 6: @h-m-m so i think we def want to ditch this commit. lmk if you want me to push a change doing that, or if it'll get rolled into the rebase

exbinary:
@h-m-m , @maebeale , can this branch be deleted?

PR #181, opened 2020-05-11
Remove actions from Public

maebeale:
- Move them to PublicPagesController, which inherits from PublicController
- Change Asks and Offers Controllers to inherit from PublicController
- Change Matches to route to contributions_public_path after completion
- Adjust routes (and change indentation)


app/controllers/community_resources_controller.rb
```diff
@@ -62,13 +53,14 @@ def community_resource_params
           :facebook_url,
           :is_approved,
           :name,
-          :organization_id,
```
maebeale, line 65: was on the fence here if this gets removed due to organization_attributes, or if both are needed ?

PR #183, opened 2020-05-11
Add ContributionsController, etc, (so we can move out of ListingsController)

maebeale:
I started to stub out moving the `contributions` data to a ContributionsController vs doing it through the ListingsController (and redirected through PublicPagesController#contributions).

@h-m-m if any of this is helpful this week or next (or the one after that! :)), please feel free to use! :)

### DIFF
--- a/spec/factories/contributions.rb
+++ b/spec/factories/contributions.rb
@@ -1,0 +1,4 @@
+FactoryBot.define do
+  factory :contribution do
+  end
+end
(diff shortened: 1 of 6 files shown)


spec/factories/contributions.rb
```diff
@@ -0,0 +1,4 @@
+FactoryBot.define do
+  factory :contribution do
```
maebeale, line 2: would we want/need this?

PR #195, opened 2020-05-11
Update references to tags

maebeale:
- Change tags in strong params and interactor -> tag_list: []
- @exbinary should i merge this, or remove the parts for ListingsController and SaveListing?

### DIFF
--- a/app/interactors/save_listing.rb
+++ b/app/interactors/save_listing.rb
@@ -2,7 +2,7 @@ class SaveListing < BaseInteractor
   string :type
   string :description, default: nil
   string :title,       default: nil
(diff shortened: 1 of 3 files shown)


app/interactors/save_listing.rb
```diff
@@ -2,7 +2,7 @@ class SaveListing < BaseInteractor
   string :type
   string :description, default: nil
   string :title,       default: nil
-  array  :tags,        default: []
+  array  :tag_list,        default: []
```
exbinary, line 5: This one i think we fixed yesterday, but good otherwise.

PR #206, opened 2020-05-12
Rails 🔄 Vue filtering api

h-m-m:
Rebase / redo of #160 

In re a comment there that I didn't respond to:

- I didn't see where publish until was implemented, so I put another date on the browse view just to make sure dates were working they way I expected — and they aren't, quite, but now I have enough info to make it right. It is easy to change this, and I'm happy to update it in another PR

Otherwise, all the follow-up suggestions I listed in #160 still apply, I suppose


app/javascript/pages/browse/ContactIcons.vue
```diff
@@ -12,7 +12,7 @@ import TagList from 'components/TagList'
 const iconNameMapping = {
   email: 'envelope',
   text: 'mobile-alt',
-  call: 'phone',
+  phone: 'phone'
```
h-m-m, line 15: I have no real preference what we name things, but I saw `phone` as a field in the database and not `call`, so I changed this temporarily so that it would work

This mapping can be long — we can put every synonym we can think of, if we want to go that far

maebeale: ah! makes sense! this is the lookup vs the dropdown

PR #207, opened 2020-05-12
Offer form bugfixes and changes to get it closer to LAMAs form

exbinary:
* Add form preamble and footer in `/offers/new.html.erb`
    * Note: this introduces some whitespace changes; diff is probably better viewed with `?w=1`
* Make field labels a bit more prominent with `is-medium`
* Add `message`s to a couple of field fields, such as name
* Give categories a label and some instructional text
* Add skills and notes fields
* Fixes bug with `preferred_contact_method` not being retained.

### DIFF
--- a/app/blueprints/person_blueprint.rb
+++ b/app/blueprints/person_blueprint.rb
@@ -2,12 +2,9 @@ class PersonBlueprint < Blueprinter::Base
   identifier :id
 
   fields :name, :email, :phone, :skills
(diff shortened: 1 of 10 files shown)


maebeale:
I went to create a new PR that does a Submission save (which generates an autoemail) after the `outcome.valid?` block, but I think I'd have to make SaveSubmission an interactor? (And, maybe should anyway since it's doing that email send?)

PR #208, opened 2020-05-13
Extract form components for reuse on Ask form

exbinary:
Several vue code refactorings that set up work on the Ask form:
* Extract `components/forms` out of `pages/offer`.
* Move `AuthTokenInput` and `DeleteButton` under forms/.
* Move `AboutSection` under `pages/orientation` since only used there.
* Fill out `components/index.js` to ease use of importing.
* Remove `PublicListings` and related code (now `/contributions`).
* Simplify `ContactFields` (see 6ccf05f).
* Extract `ErrorMessages` and `NameField` components.


app/javascript/pages/Offer.vue
```diff
@@ -85,12 +85,14 @@
...
-import LocationFields from './offer/LocationFields'
-import ServiceAreaField from './offer/ServiceAreaField'
+import {
+  AuthTokenInput,
+  CategoryFields,
+  ContactFields,
+  LocationFields,
+  ServiceAreaField,
+  SpacerField
+} from 'components/forms'
```
maebeale, line 76: pretty

spec/javascript/pages/Offer.spec.js
```diff
@@ -83,16 +84,5 @@ describe('Offer', () => {
         assert.equal($wrapper.get('input[name*="email"]').attributes('required'), 'required')
       })
     })
-
-    describe('when a contact field is updated', () => {
```
maebeale, line 87: is this spec not needed anymore?

exbinary: Yeah, it's not. Turns out i had over-complicated `ContactFields`.

This spec was testing that the parent object's value was updated by `ContactFields`, but:
> It actually doesn't need to emit update/change events up to its parent
because we don't need those updated values for anything. Those values
are stored in the corresponding inputs and submitted when the form
submits.

PR #212, opened 2020-05-13
Request support

exbinary:
Minimal, functional `/asks/new` 🐇 .

Missing some questions when compared to the LAMA [request form](https://docs.google.com/forms/d/10XnUMY4wwp4o2v0UmOQdQ6XcogjkSMtzlp-iXI65v6s/edit):
* "Do you wish to be contacted about donations..."
* "Does your request need to remain anonymous?"
* "Number of people in the household"
* "Is a member in your household over the age of 50?"
* Recurrence and Urgency, which we're planning to attach to categories.

Screen Shot 2020-05-12 at 22 40 12
Screen Shot 2020-05-12 at 22 40 44

### DIFF
--- a/app/javascript/components/forms/index.js
+++ b/app/javascript/components/forms/index.js
@@ -1,5 +1,6 @@
 import AuthTokenInput   from './AuthTokenInput'
 import CategoryFields   from './CategoryFields'
+import CommentsField    from './CommentsField'
 import ContactFields    from './ContactFields'
 import DeleteButton     from './DeleteButton'
 import ErrorMessages    from './ErrorMessages'
@@ -11,6 +12,7 @@ import SpacerField      from './SpacerField'
 export {
   AuthTokenInput,
   CategoryFields,
+  CommentsField,
   ContactFields,
   DeleteButton,
   ErrorMessages,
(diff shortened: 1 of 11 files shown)


app/javascript/components/forms/index.js
```diff
@@ -1,22 +1,26 @@
...
 import LocationFields   from './LocationFields'
 import NameField        from './NameField'
 import ServiceAreaField from './ServiceAreaField'
 import SpacerField      from './SpacerField'
+import SubmitButton     from './SubmitButton'
 
 export {
   AuthTokenInput,
   CategoryFields,
+  CommentsField,
```
maebeale, line 16: was comments not on offer form?

app/javascript/pages/Offer.vue
```diff
@@ -31,7 +31,20 @@
...
+      <p class="title is-4">
+        What are you able to offer?
+      </p>
+      <p>
+      Note: Risks and best practices will be discussed before any work is done.
+      People over 50 and people with vulnerable conditions should be especially cautious in these times,
+      so this option will not be right for everyone.
+      </p>
+      <p class="mt-1 mb-1">
+      Your safety is community safety. Do what is best.
```
maebeale, line 44: <3

app/javascript/components/forms/index.js
exbinary: It was, i just hadn't extracted it.

BTW, this is currently saving to the `description` field, which it probably shouldn't be.

maebeale: yeah, `description` is for the metadata accordion for each checkbox

i'm adding in a `SubmissionResponse` model to be able to import prior data that doesn't match our forms, so this comment could later get stored as one of those

PR #213, opened 2020-05-13
[DRAFT] Add email and log triggers

maebeale:
This is to demo the following process for now, until we add this functionality to interactors on the Vue side

- create Submission (log) from listing data + current_user
  - we want to create a non-editable log of their original form responses
- confirmation email
  - everyone who fills out a form should get a confirmation email
- communication_log
  - all autoemails sent by the system should get logged as communication_log
- redirect user to submission_thank_you_path


app/controllers/listings_controller.rb
```diff
@@ -71,6 +72,23 @@ def set_form_dropdowns
       @available_tags = Category.visible.pluck(:name) + @listing&.tag_list || []
     end
 
+    def create_submission_notification_and_log!(listing)
```
maebeale, line 75: we prob want to move this out of here

app/controllers/submissions_controller.rb
```diff
@@ -64,6 +56,16 @@ def destroy
     end
   end
 
+  def email_confirmation_and_log(submission, current_user)
```
maebeale, line 59: needs to be moved

app/models/submission.rb
```diff
@@ -13,4 +13,23 @@ def category_list
   def urgency
     "I can wait a week"
   end
+
+  def self.create_from_listing_and_log!(listing, current_user)
```
maebeale, line 17: maybe move to interactor?

maebeale:
this "works," but is not ideal. @exbinary might you have some time to pair w me this week on this (and adding relevant tests?)

exbinary:
Some pairing notes between @maebeale and @exbinary 

Functionality needed on form submission:
* send two emails
    - submitter
    - admin
* save comm log of both emails

Patterns/questions to consider:
* add error handling to email sending
* establish pattern
* add some logging (do we want to do this?)
* mailer code
    - using a class method
    - passing email object around?
    - being able to resend emails
    - use of partials
    - error handling
    - logging

exbinary:
Closing in favor of #758 and #759 .

PR #222, opened 2020-05-13
Adding update action to AsksController

maebeale:
- Since it's no longer inheriting from ListingsController, we don't get it from super


app/controllers/asks_controller.rb
```diff
@@ -18,6 +18,16 @@ def create
     end
   end
 
+  def update # for rails-side edits
```
maebeale, line 21: @exbinary is this ok w you?

PR #224, opened 2020-05-13
Submission responses spike

maebeale:


.env.ci.example
```diff
@@ -11,7 +11,6 @@ MUTUALAID_DATABASE_PASSWORD=testing123
 
 # SendGrid api key
 SENDGRID_API_KEY=123
-# SendGrid credentials
```
maebeale, line 14: including this in this PR bc it's making it easier for me to cross-compare my local env files

Gemfile
```diff
@@ -34,6 +34,9 @@ gem 'sendgrid-ruby'
 gem 'simple_form'
 gem 'webpacker', '~> 4.0'
 
+# email and sms
+gem 'sendgrid-ruby'
```
maebeale, line 38: this seems to already be in the project, but rebase isn't detecting it?

app/mailers/submission_mailer.rb
```diff
@@ -11,11 +11,11 @@ def new_submission_confirmation_email(submission)
 
     @form_name = @submission.form_name
     if @form_name.downcase.include?("ask")
-      @form_contact = current_organization.ask_form_contact
+      @form_contact = Organization.current_organization.ask_form_contact
```
maebeale, line 14: change to model method since @current_organization isn't available in the mailer

PR #236, opened 2020-05-14
Browse improvements

h-m-m:
filtersample

## Highlights

- More types of filtering
- More sensible checkbox behavior on page load
- More helpful icons
- Partial support for filtering by urgency

### Higher priority
- Allow filters changes to update the URL so that copying the current page address or reloading the page doesn't lose the filters you changed
- Make the buttons at the bottom of each contribution do something
- Implement the remaining data still commented out as "not yet implemented" in contribution_blueprint_spec.rb (availability, publish_until)

### Possible later work
- Date format parsing on js side (might be better to pass seconds since epoch from Rails and then use setTime())
- Add more tests, or break up existing tests
- Transition smoothing? Better UI when there's only one or two Tiles?

## Implementation Details

The controller takes the params, does some basic `permit` work to prevent, and reorganizes the data so that data like `{'Category' => {3 => true, 5 => true, 9 => true}}` becomes `{'Category' => [3,5,9]}`

I made a `BrowseFilter` module to then handle all the rest of turning filters into ActiveRecord scopes. It's a module at the moment and not a class because I couldn't figure out any reason why it needed to keep track of state. Also, I put it in the models folder, which feels really wrong to me. Please let me know where you think it should go.

For `BrowseFilter` I wanted to be very careful that nobody could stumble into query models we don't want them to query. I also wanted it to be extensible. In the process, I made it more complex than I'm happy with. Specifically, it's a little more complicated than I can actually understand as I'm writing this PR, so if you can't understand it either, you're not alone. This is not good, and is also a natural result of me doing less TDD than I'd like to. I suspect we'll have to redo this code at some point. 

That said, it somehow works really well. I'll be going to bed soon after I add some comments inline to the code below


app/javascript/pages/Browse.vue
```diff
@@ -27,8 +27,15 @@ export default {
...
   props: {
     contributions: {type: Array},
-    filterTypes: {type: Array},
-    initialFilters: {type: Array, default: () => []},
+    filterTypes: {type: Array, default: ()=>[]},
+    initialFilters: {
+      type: Array,
+      default: function () {
+        return [].concat(
+          ...this.filterTypes.map((fType) => fType.filters.map((filter) => filter.id))
```
h-m-m, line 35: I'm using `[].concat(...array)` as a substitute for `array.flat()`

`Array.prototype.flat` is left intact by babel/webpack in our current config, and icaniuse.com lists its lack of support at above 10%. I've been aiming for 2% or 3% so far in this project

app/javascript/pages/browse/Filters.vue
```diff
@@ -30,5 +34,25 @@ export default {
...
+  data() {
+    return {
+      hasFilterIcons: this.filterTypes
+        .map((type) => type.filters)
+        .reduce(function (memo, filters) {
+          filters.map(function (filter) {
+            memo[filter.name] = MappedIconList.iconNameMapping[filter.name]
+          })
+          return memo
+        }, {}),
```
h-m-m, line 54: I find this fairly illegible, too

app/models/urgency_level.rb
```diff
@@ -0,0 +1,21 @@
+UrgencyLevel = Struct.new(:id, :name, :other)
+# UrgencyLevel
+# We can convert this to some other type of class, such as an ActiveRecord model
+# I'm hard-coding the options for now, though
```
maebeale, line 4: 👍

app/controllers/listings_controller.rb
```diff
@@ -1,12 +1,9 @@
 class ListingsController < ApplicationController
   before_action :set_listing, only: [:show, :edit, :update, :destroy, :match, :match_confirm]
-  CONTRIBUTION_MODELS = { 'Ask' => Ask, 'Offer' => Offer }.freeze
```
maebeale, line 3: 👍

app/javascript/components/MappedIconList.vue
```diff
@@ -8,14 +8,17 @@
...
-
 const iconNameMapping = {
-  email: 'envelope',
-  text: 'mobile-alt',
-  phone: 'phone'
+  Email: 'envelope',
+  Text: 'mobile-alt',
+  Phone: 'phone',
+  Call: 'phone',
+  Ask: 'hand-sparkles',
```
maebeale, line 16: ❤️

app/javascript/pages/browse/TileListItem.vue
```diff
@@ -9,15 +9,16 @@
           <b-icon v-if="showUrgentIcon" icon="exclamation-triangle" size="is-small" />
           {{ urgency.name }}
         </b-tag>
+        <MappedIconList :contactTypes="[{id: contribution_type, name: contribution_type}]" />
```
maebeale, line 12: this is contactTypes, but using contribution_type data?

db/seeds.rb
```diff
@@ -69,7 +69,7 @@
 end
 
 [['Call', 'phone'], ['Text', 'phone'], ['Email', 'email'], ['WhatsApp', 'phone']].each do |(name, field)|
-  ContactMethod.where(ContactMethod.arel_table[:name].matches(name)).first || ContactMethod.create!(name: name, field: field)
+  ContactMethod.where(ContactMethod.arel_table[:name].matches(name)).first || ContactMethod.create!(name: name, field: field, enabled: true)
```
maebeale, line 72: 👍

app/models/browse_filter.rb
```diff
@@ -0,0 +1,38 @@
+# BrowseFilter code currently makes a lot of assumptions about the model coming in
+# I'm going to have to some lifting to make it accept other models
+# but at least all these changes will be isolated to this class
```
maebeale, line 3: this is great! i don't hate it living in models, either.

app/javascript/pages/browse/Filters.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <section>
     <h2 class="title">Filters</h2>
-    <b-collapse v-for="(type, index) of filterTypes" :key="index" :open="true">
+    <b-collapse v-for="(type, index) of filterTypes" :key="index" :open="initialOpenStatus(index)">
```
maebeale, line 4: i don't see where we set this. on my dev, only the first filter section is open by default.

app/javascript/pages/browse/Filters.vue
```diff
@@ -30,5 +34,25 @@ export default {
     prop: 'currentFilters',
     event: 'change',
   },
+  methods: {
+    initialOpenStatus(index) {
+      return index === 0
```
maebeale, line 39: maybe this is why only first is open?

app/controllers/contact_methods_controller.rb
```diff
@@ -63,6 +63,6 @@ def set_form_dropdowns
     end
 
     def contact_method_params
-      params.require(:contact_method).permit(:name, :field, :enabled)
+      params.require(:contact_method).permit(:name, :field, :enabled, :icon_class)
```
maebeale, line 66: oops. i merged into your PR. is that ok?

app/javascript/pages/browse/TileListItem.vue
h-m-m: Good point — I'll update this param name

app/javascript/pages/browse/Filters.vue
h-m-m: Yup. We can easily change the behavior here to be whatever. 

I thought this was a reasonable first go at it. All expanded looked too cluttered to me. None expanded might lead to people glossing over it and not noticing it's expandable. The first item should be most important, so that seemed an easy, reasonable start

app/controllers/contact_methods_controller.rb
h-m-m: Yup, this is fine. We should then do something that will pull this info forward into the Browse page, but that can wait for a little bit

h-m-m:
Match button info:
Screen Shot 2020-05-14 at 4 34 43 PM

h-m-m:
updated "match" link to be "respond" as per discussion with @maebeale 
Screen Shot 2020-05-14 at 5 01 52 PM

PR #246, opened 2020-05-14
update list contribution browser to be in line with latest changes

h-m-m:
## Notes
All the filtering from #236 works just fine here

Still would be nice if the site could remember whether you last used tile or list browse instead of forgetting every page load

### Largest breakpoint
Screen Shot 2020-05-14 at 5 11 22 PM


maebeale:
looks amazing! i'm pressing merge. hope is ok! :)

PR #249, opened 2020-05-16
Docs updates

maebeale:
- Refactor README
  - Remove references to Heroku CLI/Procfile since we're not using it
  - Add more info re testing
  - Try to clarify and split out deployment from Heroku/Docker/local setup stuff
- Update ORIENTATION with some additional detail
- Change license and code of conduct to reflect current Ruby for Good relationship (tentative)

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,9 +180,12 @@ You use `FactoryBot.create(:user)` before the `login_as` method and pass it in a
 BE SURE to include the line ```after { Warden.test_reset! } ``` after the `before do` block with the `login_as` method
 within it. This allows for any unexpected state data of the user from hanging around and causing errors.
 
-
 # Deploying the app
 ## If you decide to deploy to Heroku:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -154,61 +160,29 @@ Also, if you would like docker-compose to run in daemon mode (which means that i
...
+BE SURE to include the line ```after { Warden.test_reset! } ``` after the `before do` block with the `login_as` method
 within it. This allows for any unexpected state data of the user from hanging around and causing errors.
 
-Additional testing for front_end specs should make use of Capybara ```sign_in/sign_out``` Capybara methods.
-
-## App Startup Troubleshooting
-If you Receive an error when trying to run
 
-```
-$ heroku local -f Procfile.dev
```
maebeale, line 177: should we be trying to keep this heroku local stuff in here? if so, we need to add a Procfile so ppl can get past this

PR #264, opened 2020-05-16
Rake seeds redux

maebeale:
- Move seeds files and scripts into relevant folders in db dir
  - update gitignore to exclue db/gitignored_csvs*
  - Move faker gem into development/test bc we call it in seeds
  - Include basic import scripts for custom_form_questions, users, and submission_responses
    (importers will be in a separate PR)
  - Extend rake db namespace to include various options for seeds
  - Add truncate_tables and tuple_counts for use in new rake tasks
  - Create basic public_template_csvs for the 3 scripts

- Update status options in communication_log model
- Update Donation model/views w CHANNELS (and view decorators like yes_no...)
- Change ServiceArea dropdown to use constant, make , and required
- Move lists out of controller and in to model for SoftwareFeedback dropdown values
- Add dropdown values for Submission to model and view
- Add required fields to service_area factory


Gemfile
```diff
@@ -34,21 +34,17 @@ gem 'sendgrid-ruby'
...
 gem 'webpacker', '~> 4.0'
 
-# email and sms
-gem 'sendgrid-ruby'
-
 group :development, :test do
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
   gem 'binding_of_caller'
   gem 'factory_bot_rails'
+  gem 'faker'
```
maebeale, line 41: using faker seeds in development seeds files

Gemfile
```diff
@@ -34,21 +34,17 @@ gem 'sendgrid-ruby'
 gem 'simple_form'
 gem 'webpacker', '~> 4.0'
 
-# email and sms
-gem 'sendgrid-ruby'
```
maebeale, line 38: was listed twice, so deleted this while in the file

app/controllers/software_feedbacks_controller.rb
```diff
@@ -48,17 +48,9 @@ def set_software_feedback
     end
 
     def set_form_dropdowns
-      @feedback_types = ["new feature", "change", "bug"]
```
maebeale, line 51: move these to the model

app/models/service_area.rb
```diff
@@ -3,9 +3,9 @@ class ServiceArea < ApplicationRecord
   translates :name
   translates :description, type: :text
 
-  belongs_to :parent, class_name: "ServiceArea", inverse_of: :service_areas, optional: true
-  belongs_to :organization, optional: true
   belongs_to :location, class_name: "Location", inverse_of: :service_areas, foreign_key: :location_id
+  belongs_to :organization
```
maebeale, line 7: remove optional: true

PR #401, opened 2020-05-23
Update datetime on contribution json to display better in the browser

### LINKED ISSUE #331: Beautify time display on /contributions
rn we're showing the full timezone. add some vue strftime there, or send something difft through the API?

h-m-m:
### Summary

Resolves #331

#### Before

Screen Shot 2020-05-23 at 11 42 49 AM

#### After
Screen Shot 2020-05-23 at 12 00 51 PM

My local timezone is set to EDT, so the time conversion here is accurate. The time conversion occurs locally, in the browser


maebeale:
woo hoo!

PR #402, opened 2020-05-23
Refactor browse filtering

h-m-m:
### Changes
- simplify how the browse filter parser is called from the controller
- duplicate the code from the listings controller into the contributions controller
- comment out the public_contributions route as it was masking a path mentioned later in the routes file
- get all the tests to pass

### Notes
I wasn't happy with this code when I wrote it — it's too complicated for me to follow and remember what I did, and I knew that when I finished writing it. I'm glad there's a reason now for me to have a go at cleaning it up, if only partially.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,7 +11,7 @@ Rails.application.routes.draw do
   get "/about",                    to: "public_pages#about",               as: "about_public"
   get "/announcements_list",       to: "public_pages#announcements",       as: "announcements_public"
   get "/community_resources_list", to: "public_pages#community_resources", as: "community_resources_public"
-  get "/contributions",            to: "public_pages#contributions",       as: "contributions_public" # TODO remove this
+  # get "/contributions",            to: "public_pages#contributions",       as: "contributions_public" # TODO remove this
 
   resources :announcements
   resources :asks, only: [:index, :edit, :update, :new, :create]
--- a/spec/requests/contributions_spec.rb
+++ b/spec/requests/contributions_spec.rb
@@ -18,10 +18,47 @@ RSpec.describe "/contributions", type: :request do
   describe "GET /index" do
     it "renders a successful response" do
       create(:listing)
-      get contributions_public_url
-      skip # TODO - get this working instead of listingscontroller
+      get contributions_url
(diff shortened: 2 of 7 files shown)


config/routes.rb
```diff
@@ -11,7 +11,7 @@
   get "/about",                    to: "public_pages#about",               as: "about_public"
   get "/announcements_list",       to: "public_pages#announcements",       as: "announcements_public"
   get "/community_resources_list", to: "public_pages#community_resources", as: "community_resources_public"
-  get "/contributions",            to: "public_pages#contributions",       as: "contributions_public" # TODO remove this
+  # get "/contributions",            to: "public_pages#contributions",       as: "contributions_public" # TODO remove this
```
h-m-m, line 14: @maebeale I commented this out so that everything would work. Do you have a preference for having a public contributions path with one name, and a logged-in-only one with another name? 

It took me more time than I'd like to admit to figure out some tests were failing because two routes were listed for the "/contributions" path. I sure would like to see a louder warning from Rails when that happens.

spec/requests/contributions_spec.rb
```diff
@@ -18,10 +18,47 @@
...
     it "renders a successful response" do
       create(:listing)
-      get contributions_public_url
-      skip # TODO - get this working instead of listingscontroller
+      get contributions_url
       expect(response).to be_successful
     end
-  end
 
+    it 'allows asking for a specific subtype of listing' do
```
h-m-m, line 25: The tests here are copied/pasted from requests/listings_spec.rb. I'm fine with removing the duplication whenever you're ready

config/routes.rb
maebeale: Hey! hmmm. I was thinking we'd move all of these routes to the ContributionsController, tho that doesn't answer re different endpoints for logged in vs not. 

My vote would be one endpoint, but I definitely see the value of having two. My worry is that we would start to implement something we want on both all the time, and get too divergent. My leaning is that we have one for a good while, until we see reason to have two. But, open to whatever y'all think.

h-m-m: Cool, then I'll delete the commented out line for now, and then merge this

PR #405, opened 2020-05-23
Drop `autoemail` contact method

exbinary:
Replaces the `autoemail` ContactMethod with an `automated` flag on `CommunicationLog`.

Not sure `automated` is the ideal word here, open to suggestions.

Note: this doesn't migrate any existing rows in `communication_logs`, so will require seed data to be reset.


app/models/contact_method.rb
```diff
@@ -1,12 +1,12 @@
...
 
-  scope :as_filter_types, -> { enabled_public.distinct(:name) }
-  scope :autoemail, -> (boolean){ where( boolean ? arel_table[:name].lower.eq("autoemail") : arel_table[:name].lower.not_eq("autoemail") ) }
+  scope :as_filter_types, -> { enabled.distinct(:name) }
+  scope :email, -> { find_by name: 'Email' }
   scope :enabled, -> { where enabled: true }
-  scope :enabled_public, ->{ enabled.autoemail(false) }
   scope :field_name, -> (field_name){ where arel_table[:field].lower.eq(field_name) }
   scope :method_name, -> (method_name){ where arel_table[:name].lower.eq(method_name) }
+  scope :sample_one, -> { order(Arel.sql('RANDOM()')).first }
```
exbinary, line 9: Allows for `ContactMethod.sample_one` instead of `ContactMethod.all.sample`. Unfortunately AR won't let me name it `sample` :(

app/views/communication_logs/_form.html.erb
```diff
@@ -14,12 +14,13 @@
       <%= f.association :delivery_method, required: true, collection: ContactMethod.all %>
       <%= f.input :delivery_status, as: :select, required: true, collection: CommunicationLog::DELIVERY_STATUSES %>
       <%= f.input :needs_follow_up, as: :radio_buttons %>
+      <%= f.input :automated, as: :radio_buttons %> <!-- FIXME: block editing automated logs? -->
```
exbinary, line 17: Should we be able to edit auto-generated communication logs?

app/views/submissions/index.html.erb
```diff
@@ -29,6 +29,7 @@
         </td>
         <td><%= submission.urgency %></td>
         <td><%= show_button(submission.person, submission.person&.name.to_s, "fa fa-user-circle") if submission.person %></td>
+        <!-- FIXME: is the logic below correct -->
```
exbinary, line 32: I didn't quite understand the logic here?

db/seeds/dev_seeds.rb
```diff
@@ -182,29 +180,32 @@
   end
   CommunicationLog.create!(person: listing.person,
                            match: match,
-                           created_by: delivery_method&.name.include?("autoemail") ? User.first : User.all.sample,
```
exbinary, line 185: If i'm reading this correctly, `delivery_method` was always going to be `autoemail`, so the ternary here wasn't doing anything?

app/views/communication_logs/_form.html.erb
maebeale: I don't think so. Do you?

db/seeds/dev_seeds.rb
maebeale: @exbinary sorry didn't mean to resolve

maebeale: I was doing a passthrough on all the ones that got created (simulating users having created some), and was thinking that some wouldn't be auto, but, maybe I'm wrong?

app/views/submissions/index.html.erb
maebeale: add the submission show button if submission has a person ?
(i might've only needed this with fake data?)

app/views/communication_logs/_form.html.erb
exbinary: Converted to an issue.

app/views/submissions/index.html.erb
exbinary: Oh, i see. I was confused by the label "Autoemail" on the button:
Screen Shot 2020-05-23 at 20 10 54
Seems like the wrong label there?

exbinary: Moved to #407

PR #409, opened 2020-05-24
Add subcategories to Ask/Offer form

exbinary:
Also show category/subcategory descriptions when selected.


maebeale:
(We're ok w only top-level selection, right?)

exbinary:
> (We're ok w only top-level selection, right?)

On this branch, yes i think so. As we add 'metadata', things will get more complicated.

PR #416, opened 2020-05-27
Correct datetime calls

maebeale:
- Change Date.today -> Time.zone.today
- Add timezone default to UTC in config


maebeale:
are we ok defaulting to UTC?

exbinary:
Definitely good with UTC.
FYI, [`Date.current`](https://devdocs.io/rails~6.0/date#method-c-current) (and its sibling [`Time.current`](https://devdocs.io/rails~6.0/date#method-c-current)) are effectively equivalent to `Time.zone.today` (and `Time.zone.now`).

PR #420, opened 2020-05-27
Update devise workflows

maebeale:
- Bulma-ify devise views
- Add buttons to User form
- Move person select to end of nested partial
- If @user == current_user, redirect to /user/edit
- Change user name to be person&.name || email

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -16,8 +16,11 @@ class UsersController < ApplicationController
   end
 
   def edit
-    @people = Person.pluck(:id, [:name, :phone, :email])
-    set_form_dropdowns
+    if @user == current_user
+      redirect_to edit_user_registration_path
+    else
+      set_form_dropdowns
+    end
   end
 
   def create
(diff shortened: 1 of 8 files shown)


app/controllers/users_controller.rb
```diff
@@ -16,8 +16,11 @@ def new
   end
 
   def edit
-    @people = Person.pluck(:id, [:name, :phone, :email])
-    set_form_dropdowns
+    if @user == current_user
+      redirect_to edit_user_registration_path
```
maebeale, line 20: using built-in devise route here, but admins prob need to manually set someone's password?

app/views/users/_form.html.erb
```diff
@@ -5,12 +5,17 @@
     <%= f.input :email %>
   </div>
 
-  <%= f.simple_fields_for :person,  method: :patch do |ff| %>
-    <%= render "person_fields", ff: ff, f: f %>
+  <%= f.simple_fields_for :person, as: :person, method: :patch do |ff| %>
```
maebeale, line 8: still playing around w nested person, here, but skipping it for now

PR #442, opened 2020-05-29
Introduce FormObjects to simplify our nested model forms

exbinary:
Currently, we're asking Interactors to do several things:
* Capture incoming params and apply them to nested objects
* Attempt to save nested objects, handling validation, transactions and error propogation
* If invalid, double for real objects during serialization

This is evidently too many responsibilities.

The intention here is to extract FormObjects whose sole responsibility is to compose a (possibly nested) object graph and apply incoming attributes. They are not responsible for validation or persistence.

Importantly, they _always_ return ActiveModel instances. This way, we can use model instances for validation, persistence and serialization.

With this change, we end up using a hybrid of rails-way and custom handling of nested objects.

In simpler cases, it may be simpler to just use AR:NestedAttributes.

In this particular case, custom form objects were necessary because we want to share some nested objects. Eg, Person and Listings should point to the same Location while Listings and Submission point to the same Person. This is logic that NestedAttributes would not handle. OTOH, we still invoke it so that we don't have to hand-roll error message propagation code.

Note that in order to support partial updates, i've introduced the concept of `given_inputs` on top of what ActiveInteraction provides. This allows us to only apply given attributes on top of existing
objects, leaving other attributes as they are.


spec/forms/submission_form_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+
+  describe 'creating a new submission' do
+    let(:params) {{
+      form_name: 'Offer_form',
+      privacy_level_requested: 'anyone',
+      service_area: service_area.id,
+      listing_attributes: {
+        type: 'Offer',
+        tag_list: ['errands','groceries'],
+        description: 'on a quiet day i can hear her breathing',
```
maebeale, line 16: Arundhati!

spec/forms/submission_form_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+
+    it 'builds a Submission instance' do
+      expect(submission).to be_a Submission
+    end
+
+    it 'populates submission fields' do
+      expect(submission.privacy_level_requested).to eq 'anyone'
+      expect(submission.form_name).to eq 'Offer_form'
+    end
+
```
maebeale, line 43: should we validate it sets a specific person and location?

spec/forms/submission_form_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+    it 'sets the specified service area' do
+      expect(submission.service_area).to eq service_area
+    end
+
+    it 'captures whole submission json'
+
+    describe 'has_one person' do
```
maebeale, line 50: should this person and location test be under the json section only? perhaps under both?

spec/forms/submission_form_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+      it 'builds a Listing instance' do
+        expect(listing).to be_a Listing
+      end
+
+      it 'populates listing fields' do
+        expect(listing.type).to eq 'Offer'
+        expect(listing.description).to eq 'on a quiet day i can hear her breathing'
+      end
+
+      it 'populates categories on the listing' do
```
maebeale, line 96: should we also test it populates urgency level? (i think we aren't passing an urgency level at all rn, so, maybe not yet?)

spec/forms/submission_form_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+    end
+
+    describe 'on save' do
+      before { submission } # do this first to get accurate tare of record counts
+
+      it 'creates new records for itself and all nested objects' do
+        expect { submission.save }
+          .to  change(Submission, :count).by(1)
+          .and change(Location, :count).by(1)
+          .and change(Person, :count).by(1)
```
maebeale, line 116: ah. i see! nvm above

spec/forms/submission_form_spec.rb
maebeale: same for language. i wonder if we should set those like we do w contact_method_email? 

`preferred_locale_en`, and `urgency_level_within_a_week` ?

(i know this is more a refactor tho, so, sorry to muddy those waters

maebeale:
#loveit
#yesssss
#thankyouforcreatingbeauty

app/forms/listing_form.rb
```diff
@@ -1,5 +1,18 @@
-# TODO: Remove this when we replace listings_controller actions with offer/ask controller actions
-class ListingForm < Reform::Form
```
maebeale, line 2: what do we lose without using Reform anymore?

app/models/listing.rb
```diff
@@ -16,6 +16,8 @@ class Listing < ApplicationRecord
   has_many :matches_as_receiver, class_name: "Match", foreign_key: "receiver_id"
   has_many :matches_as_provider, class_name: "Match", foreign_key: "provider_id"
 
+  accepts_nested_attributes_for :location
```
maebeale, line 19: i know we've gone back and forth. 

last i understood, we were only going to put Location on Person 
(and ServiceArea, Organization, and CommunityResources (if different than Organization's Location))

open to this approach too. maybe we just go w it since it's written that way here?

spec/forms/submission_form_spec.rb
```diff
@@ -107,13 +111,38 @@
...
 
     describe 'on save' do
-      before { submission } # do this first to get accurate tare of record counts
+      context 'with valid params' do
+        before { submission } # do this first to get accurate tare of record counts
+
+        it 'creates new records for itself and all nested objects' do
+          expect { submission.save }
+            .to  change(Location,   :count).by(1)
+            .and change(Listing,    :count).by(1)
```
maebeale, line 120: should there be another test where it's 1 submission and 2+ listings?

app/javascript/components/forms/ContactFields.vue
```diff
@@ -44,8 +44,10 @@ export default {
     fieldNamePrefix: {type: String, default: ''},
   },
   data() {
+    const preferrence = this.person.preferred_contact_method
     return {
-      preferredContactMethodId: this.person.preferred_contact_method,
+      preferredContactMethodId: preferrence ? preferrence.id : null,
```
maebeale, line 49: is this a typo?

spec/javascript/pages/Offer.spec.js
```diff
@@ -35,6 +35,10 @@ describe('Offer', () => {
     person: $person,
   }})
 
+  def('submission', () => { return {
+    listing: $offer,
```
maebeale, line 39: is this $offer implicit bc of import Offer ?

spec/forms/submission_form_spec.rb
exbinary: I'm planning to add those as we go from one listing to listing-per-category.

spec/forms/submission_form_spec.rb
exbinary: Same as above.

spec/forms/submission_form_spec.rb
maebeale: yup re in both -- json is absolute record of initial submission, assns help us run reports later, and navigate (and at some point, possibly manage) collections of listings.
 
and, sounds great

app/models/listing.rb
maebeale: cool. i say let's add it now, and if we remove later, great. much harder to add data after the fact than to trim it down! :)

maebeale: having it on listing will help w running queries to plan dispatcher drops, so i like it this way too...

app/controllers/offers_controller.rb
```diff
@@ -1,29 +1,35 @@
 class OffersController < PublicController
```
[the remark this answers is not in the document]
maebeale: (and then we'd add Listing Update to ListingsController ?)

app/forms/base_form.rb
```diff
@@ -0,0 +1,9 @@
+class BaseForm < ActiveInteraction::Base
+  def self.build params
+    run! params
```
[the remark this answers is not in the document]
maebeale: TY. VERY helpful info.

app/forms/person_form.rb
```diff
@@ -0,0 +1,17 @@
...
+    record  :preferred_contact_method, class: ContactMethod
+    record  :location
+    string  :name
+    string  :email
+    string  :phone
+    string  :skills
+  end
+
+  def execute
+    (id? ? Person.find(id) : Person.new).tap do |person|
```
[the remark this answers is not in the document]
maebeale: ooo. i like that. either way good by me tho. sometimes redundancy isn't terrible.

app/javascript/components/forms/ContactFields.vue
exbinary: Ah yes, good catch!

spec/javascript/pages/Offer.spec.js
exbinary: There's a `def('offer' ...` a couple lines above, probably was collapsed in the diff. This is how `bdd-lazy-var` does the equivalent of rspec's `let` in JS specs.

maebeale:
Ah. I clicked view file and didn't see it, but didn't open in my console
before writing that.

PR #444, opened 2020-06-03
Add location type to ask/offer form

### LINKED ISSUE #191: Add dropdown of location type to Ask & Offer form
replace currently hard-coded field

exbinary:
Also introduces a `ConfigurationBlueprint` object that consolidates `ServiceArea`, `LocationType`, `ContactMethod` configurations so controllers don't have to know this logic.

Fixes #191

### DIFF
--- a/app/blueprints/configuration_blueprint.rb
+++ b/app/blueprints/configuration_blueprint.rb
@@ -1,0 +1,13 @@
+class ConfigurationBlueprint < Blueprinter::Base
+  association :categories, blueprint: CategoryBlueprint, view: :with_subcategories do
+    Category.visible.roots.includes(:categories)
+  end
+
+  association :contact_methods, blueprint: ContactMethodBlueprint do
+    ContactMethod.enabled
+  end
+
+  association(:service_areas, blueprint: ServiceAreaBlueprint) do
+    ServiceArea.all.eager_load(:string_translations, :text_translations)
+  end
+end
--- a/app/blueprints/configuration_blueprint.rb
+++ b/app/blueprints/configuration_blueprint.rb
@@ -7,6 +7,10 @@ class ConfigurationBlueprint < Blueprinter::Base
     ContactMethod.enabled
   end
 
+  association :location_types, blueprint: LocationTypeBlueprint do
+    LocationType.all  # FIXME: fix seeds so we can change this to `.visible`
+  end
+
   association(:service_areas, blueprint: ServiceAreaBlueprint) do
     ServiceArea.all.eager_load(:string_translations, :text_translations)
   end
--- a/app/blueprints/configuration_blueprint.rb
+++ b/app/blueprints/configuration_blueprint.rb
@@ -12,6 +12,6 @@ class ConfigurationBlueprint < Blueprinter::Base
   end
 
   association(:service_areas, blueprint: ServiceAreaBlueprint) do
-    ServiceArea.all.eager_load(:string_translations, :text_translations)
+    ServiceArea.publicly_visible.eager_load(:string_translations, :text_translations)
   end
 end
--- a/app/javascript/components/forms/LocationFields.vue
+++ b/app/javascript/components/forms/LocationFields.vue
@@ -5,26 +5,53 @@
       label="Street address"
       custom-class="is-medium"
     >
-      <b-input :name="withPrefix('street_address')" :value="street_address" />
+      <b-input :name="withPrefix('street_address')" v-model="streetAddress" />
     </b-field>
 
-    <!-- TODO: replace this with a dropdown -->
-    <b-input :name="withPrefix('location_type')" type="hidden" value="1" />
+    <b-field
+      v-if="streetAddress.length"
+      :label-for="withPrefix('location_type')"
+      label="Address type"
+      custom-class="required-field is-medium"
+    >
+      <b-select
+        :name="withPrefix('location_type')"
+        :value="location_type.id"
+        placeholder="Type"
+        required
+      >
+        <option v-for="{id, name} in location_types" :key="id" :value="id" >
+          {{ name | replace('_', ' ') | capitalize }}
+        </option>
+      </b-select>
+    </b-field>
   </div>
 </template>
 
 <script>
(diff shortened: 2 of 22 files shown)


app/blueprints/configuration_blueprint.rb
```diff
@@ -0,0 +1,17 @@
+class ConfigurationBlueprint < Blueprinter::Base
```
maebeale, line 1: Love this

app/javascript/components/forms/LocationFields.vue
```diff
@@ -5,26 +5,53 @@
...
+      v-if="streetAddress.length"
+      :label-for="withPrefix('location_type')"
+      label="Address type"
+      custom-class="required-field is-medium"
+    >
+      <b-select
+        :name="withPrefix('location_type')"
+        :value="location_type.id"
+        placeholder="Type"
+        required
```
maebeale, line 21: Is this only required if a location is present?

app/javascript/utils/string.js
```diff
@@ -2,3 +2,8 @@ export function capitalize(string) {
   if (!string) return ''
   return string.charAt(0).toUpperCase() + string.slice(1)
 }
+
+export function replace(string, target, replacement) {
+  if (!string) return ''
```
maebeale, line 7: Does !string detect empty string?

app/blueprints/configuration_blueprint.rb
```diff
@@ -0,0 +1,17 @@
...
+  association :categories, blueprint: CategoryBlueprint, view: :with_subcategories do
+    Category.visible.roots.includes(:categories)
+  end
+
+  association :contact_methods, blueprint: ContactMethodBlueprint do
+    ContactMethod.enabled
+  end
+
+  association :location_types, blueprint: LocationTypeBlueprint do
+    LocationType.all  # FIXME: fix seeds so we can change this to `.visible`
```
maebeale, line 11: Yup

app/blueprints/configuration_blueprint.rb
```diff
@@ -0,0 +1,17 @@
+class ConfigurationBlueprint < Blueprinter::Base
+  association :categories, blueprint: CategoryBlueprint, view: :with_subcategories do
+    Category.visible.roots.includes(:categories)
+  end
+
+  association :contact_methods, blueprint: ContactMethodBlueprint do
+    ContactMethod.enabled
```
maebeale, line 7: I'm starting to want the `enabled` to be renamed to `visible` or `display_to_public` or vice versa so they're all the same. Display to public is clearer to me the more I go in and out of the codebase, but, not sure where you're at in this rn? (I know it doesn't/ shouldn't affect this PR)

app/javascript/components/forms/LocationFields.vue
maebeale: address is empty, but still getting location_type required ?

exbinary: Oops! Good catch. Pushed up a [commit](26b84ef) to accommodate empty location. Also threw in that [`find_or_new`](39067e4) method :).

app/javascript/utils/string.js
exbinary: Yep.
```js
!null //=> true
!''   //=> true
!'s'  //=> false
```

app/blueprints/configuration_blueprint.rb
exbinary: I don't have a strong preference on any of these options and agree consolidating on one would be good. `display_to_public` sounds like a verb to me instead of an adjective, so something like `publicly_visible`? But really don't mind either way.

app/blueprints/configuration_blueprint.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  association :location_types, blueprint: LocationTypeBlueprint do
+    LocationType.all  # FIXME: fix seeds so we can change this to `.visible`
+  end
+
+  association(:service_areas, blueprint: ServiceAreaBlueprint) do
```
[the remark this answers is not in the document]
exbinary: Probably. Should i roll that in here?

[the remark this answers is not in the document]
maebeale: my vote would be yes, but def defer to you

app/blueprints/configuration_blueprint.rb
maebeale: ah. yup. like `publicly_visible`
i think we should do it in a separate PR tho, do you?

exbinary: Agreed.

app/blueprints/configuration_blueprint.rb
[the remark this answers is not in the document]
exbinary: Cool, easy enough to roll in. As i'm type out `.publicly_visible` though, i wonder if just `.public` is descriptive enough?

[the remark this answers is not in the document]
maebeale: maybe? i like including the visible part better, bc public is a word that sometimes i think means railsy things, but, yeah. it def does work. 

i'd love to keep the same name in code as in the db and what's presented to users, if possible, bc i think this is going to be a point that requires a hint and/or training. i know we'll need to go to i18n anyway, but for the english side to align i think would help w things.

def will defer to you. `public` does work.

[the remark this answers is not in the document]
exbinary: Well, we need to go through and consolidate all this between different models and the DB, so we'll have a chance to make a final decision. I'll name it `public` for now and maybe by then we'll have a more solid opinion :)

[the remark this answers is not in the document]
exbinary: Whomp! Yeah, AR won't let me name it `public` 😄 .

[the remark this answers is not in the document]
exbinary: Pushed as `publicly_visible`. Other synonyms for when we come back through:
 known, published, publicized, in circulation, exposed, overt, accessible, available.

PR #448, opened 2020-06-04
Change mobility to explicitly convert to type: string

maebeale:


maebeale:
> Looks good. It appears we also have the [default](https://github.com/maebeale/mutual-aid/blob/a00e5ce17fb8c91a11af5349aadcc88568ad6f05/config/initializers/mobility.rb#L89) set to string, so an alternative could be to drop the explicit types everywhere.

That's what I thought, too, and had added default and removed from models. But the instructions say to explicitly list. 

I think the default maybe needs to come out bc it doesn't seem to be working reliably. I think this is part of why some were being created and some weren't. Going to dig in to that a little more, too.

PR #451, opened 2020-06-04
Hotfix: drop text_translations reference in ConfigurationBlueprint

exbinary:
The association was dropped from `ServiceArea` as a result of a00e5ce.
Ask and Offer forms are broken w/o this fix.


maebeale:
i think the removal was a mistake, but, yeah, ok for now

Issue #61, opened 2020-04-16
Add Pundit


maebeale:
There is an existing PR for this that needs to be finished. Please reach out to a team lead if interested! 

* Need 4 levels for now: neighbor, volunteer, admin, sys_admin
  - Display variations for these levels are not yet decided
  - `user.sys_amin?` will need to replace `user.admin?` on users/form so most users aren't able to delete accounts
* Consider removing/replacing @admin_status declaration and usage (but maybe this can be a separate issue? I would love to be able to use that var (A LOT) while developing.)

armahillo:
What is this blocked by?

I've not set up `pundit` before but I've managed its configuration -- this'll be a fun exercise. I'll configure the stuff in this issue, but see if I can skip locking things down in the display. I *think* that can be deferred?

maebeale:
I noticed that block this morning. I don't think anything is blocking it at this point!

armahillo:
OK cool! I'll carry it as far as I can with the info in the issue desc

maebeale:
* Neighbor is to help us get on track for peer-to-peer
* volunteer is unclear, but essentially is someone who would drop things off to someone else who needed groceries, for ex
* personally, I think there should be a dispatcher role where they can edit listings
* admin can see donations, finances
* sys_admins can edit system_settings, categories, system_locales, contact_methods, etc

maebeale:
Those notes are more role based than permission based. Prob we need a hybrid?

armahillo:
If you want to do Role based permissions, `cancancan` is good for that. 

Let me spike out an attempt with Pundit and see if I run into any friction, though.

maebeale:
Yay! Thanks! We've needed to get Pundit going for a while now. 

Fwiw, from prior conversations, we def prefer pundit over cancancan (and I have pundit experience but limited cancancan.), but open to your findings!

armahillo:
@maebeale Does the entire site require authorization except for the public pages? Or are there some actions that don't require being signed in? (This might be apparent from looking at the controllers, I was just wondering if you knew offhand)

maebeale:
I was trying to design for all public in public controller, but that's not where we ended up. 

There are a few endpoints in other controllers that don't require authorization

armahillo:
OK, thanks! :+1:

maebeale:
Contributions/ listings/ asks/ offers/ announcements/  community_resources all have publicly accessible endpoints (pretty sure).

[...]

PR #464, opened 2020-06-05
Adds Roles to Users (Preparation for adding Pundit)

armahillo:
Ref #61 (additional work is necessary to add `Pundit`)

> Please review this thoroughly. I am only superficially acquainted with the app, and although I was as meticulous as possible (see below), it is entirely possible I have missed something and I don't want to inadvertently create problems. This PR touches a mission-critical model (`User`) as well as also making database changes. 

# Summary of work
Pundit doesn't implicitly add roles. We have to add those separately. This PR solely focuses first on adding the roles necessary. In the interest of keeping parity, I've searched through the codebase and done find/replacing on existing methods (`:admin?` and `:sys_admin?`) and replaced them with the new hotness (`:admin_role?` and `:sys_admin_role?` respectively). 

As we were not doing any kind of role-based authorization (mediated through the model) I believe these changes will be transparent -- the end-result will be all users having `:unset` as their  `:role` until it is manually overridden. The YML-based / params-based admin-settings will still work, as before.

## Changes to the database
~The big honkin'~ **An important** change here is a database migration, **a string column called `role`**. ~This creates a PG Enum type called `user_role` and then creates a `role` column of that type. **If you want to add additional types in the future, you will need to modify this type**. That's an unfortunate tradeoff for the benefit.~ By using a ~custom enum~ **string** type, this means that the records in the database table will be human-readable strings that are explicit (eg. `admin`, `neighbor`) instead of integer values, which is typical of Rails enums (`1`, `2`). ~While you *could* use strings for the column type, this method is....better....magically.~  I also added an index for this column. The migration is safely bi-directional; I ran it in both directions multiple times and it's solid. There is an additional role type called `'unset'` which is the default. This seemed safest.

~Using a custom type means that the schema can't default to the pretty rails `schema.rb` and instead must use raw SQL. This change was also made. Unless you're the type of person who likes to read through a `schema.rb` file for kicks, you likely won't notice this difference; you can use `schema:load` and all that just like before.~

## Changes to the User model
I found a good reference for a robust solution in adding Enums. I really like their approach. It does the standard `enum` directive, but sets it up as a hash (this is a transparent change) -- this allows the value to be stored as a string instead of an integer -- and also appends a `_role` suffix -- this is why there was some necessary finding and replacing. I removed `admin?` and `sys_admin?` method since they are superfluous now -- instead you'll use `admin_role?` and `sys_admin_role?`. You can **also** use `neighbor_role?` or `dispatcher_role?` etc. (I added dispatcher because why not). As with any normal Rails enum usage, the magic scopes (eg. `User.neighbor_role`, `User.admin_role` etc) are all created. I verified they work in the rails console.

## Changes to the factories / tests
We previously had an `:admin` alias for `:user` factory. I removed the alias and replaced it with several traits -- one for each role. Now, instead of doing `create(:admin)` you would do `create(:user, :admin)`. If you really want to do `create(:admin)` instead, you can add a subfactory (I personally prefer to do traits so that it's clearer which object is being created). This was only used in one place (controller macros) and that's been updated as well.

## Other changes to the views and controllers
Any place where we were checking `current_user.admin?` or `current_user.sys_admin?`, it's been replaced with the new version of this, noted above. There were several places.

Eventually, these will be replaced entirely with pundit policies, but I felt it would be more prudent to do a smaller PR that achieved parity with current function. 

# Additional Work to be done in production
If you know the designated roles of various users, you'll want to set them after this is deployed and the `users` table updated to have `:role`. Every user will default to `:unset` as the type. Once `Pundit` is added, later, this will potentially lock out users.

If you want to change the default role, you can do that in a migration like normal; default isn't set anywhere else.

# Testing
The specs run clean (though we need more spec coverage so this doesn't mean a whole lot right now). I tested it manually and it works OK.

# References
I used this reference for creating the enum type for Role. I like the approach they take, it's robust and flexible: https://naturaily.com/blog/ruby-on-rails-enum -- we did NOT do the PG Enum column type, but did a string instead.


armahillo:
CI is failing because of this error:

```
+ bin/rails db:setup
/home/circleci/project/.bundle/ruby/2.7.0/gems/activerecord-6.0.2.2/lib/active_record/type.rb:27: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/circleci/project/.bundle/ruby/2.7.0/gems/activerecord-6.0.2.2/lib/active_record/type/adapter_specific_registry.rb:9: warning: The called method `add_modifier' is defined here
/home/circleci/project/.bundle/ruby/2.7.0/gems/actionpack-6.0.2.2/lib/action_dispatch/middleware/stack.rb:37: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/circleci/project/.bundle/ruby/2.7.0/gems/actionpack-6.0.2.2/lib/action_dispatch/middleware/static.rb:110: warning: The called method `initialize' is defined here
/home/circleci/project/.bundle/ruby/2.7.0/gems/activemodel-6.0.2.2/lib/active_model/type/integer.rb:13: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/circleci/project/.bundle/ruby/2.7.0/gems/activemodel-6.0.2.2/lib/active_model/type/value.rb:8: warning: The called method `initialize' is defined here
/home/circleci/project/.bundle/ruby/2.7.0/gems/activerecord-6.0.2.2/lib/active_record/connection_adapters/postgresql/oid/specialized_string.rb:12: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/circleci/project/.bundle/ruby/2.7.0/gems/activemodel-6.0.2.2/lib/active_model/type/value.rb:8: warning: The called method `initialize' is defined here
Database 'mutualaid_test' already exists
rails aborted!
failed to execute:
psql -v ON_ERROR_STOP=1 -q -X -f /home/circleci/project/db/structure.sql mutualaid_test

Please check the output above for any errors and make sure that `psql` is installed in your PATH and has proper permissions.
```

I am not familiar with CircleCI but this appears to be a configuration issue. It's failing because it's trying to execute the SQL (see above, I had to change how the schema is stored). I'm not sure how to fix this -- This URL looks promising: https://discuss.circleci.com/t/using-psql-command-in-circleci-2-0-build/14785 -- it's late and I want to go to bed :) Can experiment with this tomorrow.

maebeale:
@armahillo you're a mensch!!!!! AWESOME work!!!! TOTALLY agree with this interim step to pundit. THANK YOU for going so deeply in to the app!!!!!

maebeale:
(p.s. I actually do read and reference schema.rb -- on a daily basis, lol. Just saying that so you know there are people like me in the world. :)  I have an app that's been in prod for 6 years and still successfully uses migrations and has an active schema.rb. my colleagues on this project already think structure.sql was inevitable, so, this is where I think we'd have ended up anyway.)

maebeale:
I am so hesitant to say anything, bc I don't want to communicate anything other than awe and gratitude for this work (!!!), but, why `enum` and not `json`, hstore, etc? We have a couple of those in the db already that CI didn't hate. If we aren't able to move the dial, maybe we could keep all that you've set up and change the data type only? (But maybe I don't understand all the magic that `enum` is giving us as far as scopes, etc.)

Happy to pair some on this if that's be preferable and/or useful. 

Soooooo excited and grateful for your deep dive -- and for your you.

armahillo:
> I am so hesitant to say anything, bc I don't want to communicate anything other than awe and gratitude for this work (!!!), but, why `enum` and not `json`, hstore, etc? We have a couple of those in the db already that CI didn't hate. If we aren't able to move the dial, maybe we could keep all that you've set up and change the data type only? (But maybe I don't understand all the magic that `enum` is giving us as far as scopes, etc.)
> 
> Happy to pair some on this if that's be preferable and/or useful.
> 
> Soooooo excited and grateful for your deep dive -- and for your you.

JSON datatype is good when the values are going to be more varying. Using an `enum` enforces a limited set. The `enum` keyword also adds some magic methods that are convenient -- there's a predicate for each key (`User.first.admin_role?`) and a scope (`User.admin_role`) automatically added. Enums are also just a common pattern for doing roles since you typically decide on the list and then don't change them.

I also have a tendency to try to use the Rails way where possible, just because it usually makes it a little less painful. Doing the PG "enum" column type is a deviation, granted, so if that would be a dealbreaker and you'd prefer to use a string value or something, I'm happy to sort that out. The default is typically to put an integer value in the column, which is basically meaningless if you ever look at table dumps (plus as you might imagine, it makes the data integrity of that column a bit risky if you change the order of the fields :grimacing: )

maebeale:
Thanks! Yeah, not usually a fan of enums bc of it binds the db to the code. But, I'm ok w going forward as you've set it up as long as the others are!

armahillo:
OK I found a compromise -- I removed the PG SQL solution and replaced it with a plain string column; because we're using a value object to mediate the enumeration, and because that was already doing it as a hash, there was no additional change required.

Two benefits:
- we can now use schema.rb again instead
- CI now passes with no additional configuration required :+1:

PR #477, opened 2020-06-08
Add Form CRUD

maebeale:
Completes #476 

Add Form model and CRUD to work up to header/footer editable, connected custom_form_questions, etc

Fields:
- contribution_type_name (since ContributionType doesn't have a table yet)
- name
- header_html
- footer_html
- display_categories:boolean (in case they want to make forms that aren't re contributions)
- organization (for shared/multi-tenant setups)


maebeale:
@exbinary yup, this is scaffolded. will push the 2nd branch too, that connects these to custom_form_questions

my vote is sanitize on view. do you have a different recommendation?

PR #487, opened 2020-06-10
Turn parent categories into accordions on Ask/Offer forms

exbinary:
Just a UI change for now, but sets us up to append add-on fields to subcategory checkboxes.

Note that the form will append the parent category tag to the submitted tags if any subcategory is selected.

Any parent category with no subcategories will show a faux subcategory with the same name, id as the parent.

Screen Shot 2020-06-10 at 00 25 07

### DIFF
--- a/spec/javascript/components/forms/CategoryFields.spec.js
+++ b/spec/javascript/components/forms/CategoryFields.spec.js
@@ -18,50 +18,53 @@ describe('CategoryFields', () => {
     }
   })
 
-  function isChecked(value) {
-    return $wrapper.find(`input[value="${value}"]`).element.checked
-  }
-
-  describe('top-level categories', () => {
-    def('categories', () => {
-      return [
-        {name: 'meals',   description: 'on wheels'},
-        {name: 'housing', description: 'is a human right'},
+  def('categories', () => {
+    return [{
+      id: 'meals',   // ids should be integers; using strings to make spec selectors more obvious
+      name: 'meals',
+      description: 'on wheels',
+      subcategories: [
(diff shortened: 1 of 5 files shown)


spec/javascript/components/forms/CategoryFields.spec.js
```diff
@@ -18,50 +18,53 @@ describe('CategoryFields', () => {
...
+      description: 'on wheels',
+      subcategories: [
+        {id: 'prepare', name: 'prepare'},
+        {id: 'deliver', name: 'deliver'}
       ]
-    })
+    }, {
+      id: 'housing',
+      name: 'housing',
+      description: 'is a human right',
```
maebeale, line 33: ❤️

spec/javascript/components/forms/CategoryFields.spec.js
```diff
@@ -18,50 +18,53 @@ describe('CategoryFields', () => {
...
+      'meals',
+      'prepare',
+    ]
+  })
 
-    it('renders checkboxes with selected tags checked', () => {
-      assert.isTrue(isChecked('meals'))
-      assert.isFalse(isChecked('housing'))
-    })
+  it('renders accordions for each parent catogory', () => {
```
maebeale, line 45: really nitpicky: catEgory?

maebeale:
I think the categories question text is now bigger than the other q's tho?
Screen Shot 2020-06-10 at 7 54 01 PM

exbinary:
> I think the categories question text is now bigger than the other q's tho?

Good catch. Fixed.

PR #504, opened 2020-06-12
Hide navbar on forms, public_controller indexes, and thank you

maebeale:
- hide_navbar for forms, public_controller indexes, and form submission thank_you page
- show "Return to admin" if user signed in

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -3,6 +3,8 @@ class AnnouncementsController < ApplicationController
   before_action :authenticate_user!, except: [:new, :create]
   before_action :set_announcement, only: [:show, :edit, :update, :destroy]
 
+  layout "without_navbar", only: [:new, :create, :show]
+
   def index
     @announcements = Announcement.order(created_at: :desc)
   end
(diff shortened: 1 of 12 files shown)


maebeale:
@exbinary you ok w this solution?

maebeale:
A separate help?

maebeale:
I was thinking yes re nav. Do you agree?

exbinary:
> A separate help?

Oops, sorry! "Would a separate _layout_ help ..."

exbinary:
> I was thinking yes re nav. Do you agree?

I think that makes sense for all our current groups where the forms are the only things `People` see. I could see a case in the future, maybe with p2p, where more of the app is visible to the public, including some navigation; in that case, it would be better to have the form use the same skin. But we can solve for that when we have to.

maebeale:
Yes, but I thought you didn't want to do it that way. ? Maybe I'm
misremembering.

maebeale:
k!

maebeale:
will add layout back in

maebeale:
k. have added back in the layout we used to have in there called "public" -- bc there are public-facing pages where we do want to have the navbar -- and instead named it without_navbar.

this seem good, @exbinary ?

PR #509, opened 2020-06-12
first pass of ansible support for mutual aid app

bascibaran:
Added ansible playbooks to deploy the db server and the app server + quickstart instructions

Also appended a line to the global `.gitignore` to ignore vim `.swp` files.


maebeale:
WOO HOOO!!!

PR #513, opened 2020-06-14
Improve readme setup instructions

compwron:
Just a minor change from getting myself setup- `rake dev:setup` doesn't exist but `rake db:setup` does so I assume that's what goes here. The env variables are needed to get `db:setup` to run because otherwise `seed.rb` fails. 

I added the gitignore lines because of my local dev setup

Add a badge for the CI build to make it easier to find on the main page

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,8 @@ Then, to run the app locally,
 ```
 $ bundle install
 $ yarn install
-$ SYSTEM_EMAIL=example@example.com SYSTEM_PASSWORD=12345678 rake db:setup
+$ # in .env, add SYSTEM_EMAIL and SYSTEM_PASSWORD, or:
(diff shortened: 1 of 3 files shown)


maebeale:
`rake dev:setup` should've been changed to: `rake db:rebuild_and_seed` or `rake db:rebuild_and_seed_dev` (we list is later but forgot that was up there)

maebeale:
AWESOME! Thanks for going through this AND for making it easier for others!!!

README.md
```diff
@@ -1,3 +1,5 @@
+[![CircleCI](https://circleci.com/gh/maebeale/mutual-aid.svg?style=svg)](https://circleci.com/gh/maebeale/mutual-aid)
```
maebeale, line 1: AWESOME! i didn't even know this was a thing!

PR #514, opened 2020-06-14
Adding Pundit, applies to User and Person models

armahillo:
Will update this shortly.


app/controllers/people_controller.rb
```diff
@@ -1,4 +1,5 @@
 class PeopleController < ApplicationController
+  before_action :authenticate_user!
```
maebeale, line 2: isn't this already in ApplicationController?

app/controllers/people_controller.rb
```diff
@@ -1,4 +1,5 @@
...
   before_action :set_person, only: [:show, :edit, :update, :destroy]
 
   def index
```
maebeale, line 5: should this have `authorize @people`?

app/controllers/users_controller.rb
```diff
@@ -16,16 +16,17 @@ def new
...
   def edit
+    authorize @user
     if @user == current_user
       redirect_to edit_user_registration_path
     else
       set_form_dropdowns
     end
   end
 
+  # NOTE: I'm not entirely sure this is used at all.
```
maebeale, line 27: yeah, i'm considering a new controller so admins can do create user+person in one step and removing this from here bc it's not used/useable

app/policies/person_policy.rb
```diff
@@ -0,0 +1,24 @@
+class PersonPolicy < ApplicationPolicy
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
+  def edit?
```
maebeale, line 9: we get this from ApplicationPolicy? not sure what standard is for adding to child policies

app/policies/user_policy.rb
```diff
@@ -0,0 +1,24 @@
...
+    @user = user
+    @record = record
+  end
+
+  def edit?
+    update?
+  end
+
+  def update?
+    user == record || user.sys_admin_role? || user.admin_role?
```
maebeale, line 14: should users bc able to edit their own roles tho? maybe dropdown is only their permission level and lower?

app/views/errors/403.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>Sorry, you aren't authorized to do that.</h1>
+<p>If you think this is wrong, please contact your administrator.</p>
```
maebeale, line 2: 🙌

app/views/users/show.html.erb
```diff
@@ -1,15 +1,25 @@
-<div class="title is-3">Profile</div>
-<div class="title is-5"><%= @user.person&.name %></div>
+<div class="title is-3">Account Settings</div>
```
maebeale, line 1: 👍

config/locales/en.yml
```diff
@@ -51,3 +51,5 @@ en:
       one: Match Feedback
       other: Match Feedback
       zero: Match Feedback
+  pundit:
+    default: 'You cannot perform this action.'
```
maebeale, line 55: totally! wtg w the details!

spec/requests/people_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+      it_behaves_like "redirects without authorization"
+      it_behaves_like "disallowed behavior as", [:user, :with_person]
+      it_behaves_like "successful with admin user authorization"
+    end
+  end
+
+  describe "POST /create" do
+    subject { post people_url, params: { person: valid_attributes } }
+
+    it_behaves_like "basic user authorization required"
```
maebeale, line 60: cool! lsn!

spec/requests/people_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+
+    context "when the user already has a person record for themselves" do
+      let(:self) { create(:user) }
+      subject { post people_url, params: { person: valid_attributes.merge(user_id: self.id) } }
+
+      before(:each) { create(:person, user_id: self.id) }
+
+      # TODO: What happens when the user already has a record created for themselves
```
maebeale, line 68: 🤷‍♀️

spec/requests/people_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+    end
+
+    context "when attempting to create a person for someone else's user ID" do
+      let(:other_user) { create(:user) }
+      subject { post people_url, params: { person: valid_attributes.merge(user_id: other_user.id) } }
```
maebeale, line 73: 👍

spec/support/basic_authorization_requirements_shared_examples.rb
```diff
@@ -0,0 +1,54 @@
+RSpec.shared_examples 'redirects without authorization' do
```
maebeale, line 1: 💃 awesome!

app/controllers/application_controller.rb
```diff
@@ -27,8 +30,24 @@ def default_url_options
...
-  def user_not_authorized
-    flash[:alert] = "You are not authorized to perform this action."
-    redirect_to(request.referrer || root_path)
+  
+  def user_not_authenticated(exception)
+    flash[:error] = "This requires authentication, please sign-in first."
+    respond_to do |format|
+      format.html { render 'devise/sessions/new.html.erb', :layout => "application", :status => 401 }
+      format.xml  { head 401 }
+      format.any  { head 401 }
```
maebeale, line 39: i think we need an `end` here

maebeale:
ack! didn't notice the `draft` tag. my b!!!

app/controllers/people_controller.rb
armahillo: Yeah -- whoops! Gotta remove that :(

app/controllers/people_controller.rb
armahillo: Maybe -- I'm still learning how to setup Pundit

app/controllers/application_controller.rb
armahillo: ah farts, you're right

app/controllers/people_controller.rb
armahillo: Looks like it! I can take this one out

app/controllers/people_controller.rb
armahillo: *shrug* Maybe? We don't really have a lot of info about the expectations of how this stuff works so I wasn't sure how the `/people` route was intended to be used. Should it be locked down?

app/policies/person_policy.rb
armahillo: I think originally I had changed it to something else and was just changing it back here. I realize it's redundant, but it also explicitly declares the expected behavior (were `ApplicationPolicy` to be changed later, that would create problems here). My personal preference for stuff like this is to have the default behaviors as a safety net but to explicitly define things that are definitely used. 

I'm happy to defer to yours / the team's preferences on this though!

app/policies/user_policy.rb
armahillo: yeah I think the controller layer is the right place to set that

spec/requests/people_spec.rb
armahillo: haha 😆 

This could be decided on later, it was just something that came to mind while I was doing it

exbinary:
This is looking great @armahillo ❤️💥.

One thought on testing: i wonder if it would be sufficient to directly test policies (`spec/policies/...`) and verify controller coverage by getting pundit [to ensure authorization happened](https://github.com/varvet/pundit#ensuring-policies-and-scopes-are-used).

I'm guessing it will be easier to test policy behavior w/o having to go through the controller and could also lead to less duplication/boiler-plate across request specs (most controllers will behave exactly the same based on whether the policy throws or not).

app/controllers/people_controller.rb
bhaibel: IME, authorizing a collection isn't very useful, but verifying that the collection was generated using a policy scope is. The best way to do that is to let the `verify_policy_scoped` after_action catch it if you forget.

app/controllers/users_controller.rb
bhaibel: I can't get the tests for it to run,  or even route to it so that I can insert a byebug! So I'm going to remove it.

spec/support/shared_contexts/authorization.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:acting_user) { FactoryBot.create(:user) }
+end 
+
+RSpec.shared_context "acting as sysadmin" do
+  let(:acting_user) { FactoryBot.create(:user, :sys_admin) }
+end
+
+RSpec.shared_context "acting as admin" do
+  let(:acting_user) { FactoryBot.create(:user, :admin) }
+end
```
exbinary, line 11: Have to admit, i've grown to dislike globally shared contexts, particularly those that define `lets` referenced in specs. I find the cognitive load involved in jumping back and forth between spec, local context, shared context and sometimes other connected contexts to be quite heavy.

I really like the local ones in `user|person_policy_spec` because the context names provide more meaning that the lets therein. These ones definitely also provide meaning, but i'm finding it difficult have certainty just from the names exactly what they do. Perhaps this is case of needing to get used to the terminology ('acting-as', 'signed-in', etc) and once i do, my mind won't want to jump around. OTOH, i wonder if inlining these ends up being easier to grok, even at the cost of some more duplication?

The ones below seem widely used in request specs so probably worth their weight.

app/policies/application_policy.rb
```diff
@@ -0,0 +1,67 @@
...
+  def new?;      add?;     end
+  def update?;   change?;  end
+  def edit?;     change?;  end
+  def destroy?;  delete?;  end
+  
+  # We default all permissions to false, and expect you to override as needed.
+  private def read?;    false;  end
+  private def add?;     false;  end
+  private def change?;  false;  end
+  private def delete?;  false;  end
```
exbinary, line 66: 👍

exbinary (commented):
This is looking great!! 🙏🏾 🙌🏾
A few thoughts and questions inline.

exbinary:
@bhaibel , i noticed this branch was showing some minor conflicts against `main`.
I went ahead and merged it in, resolving conflicts, with the hope this is more helpful than annoying! 🤞🏾😅

app/policies/person_policy.rb
```diff
@@ -0,0 +1,44 @@
...
+  def change?
+    person_attached_to_acting_user? ||
+      sys_admin? ||
+      admin?
+  end
+
+  def add?
+    person_attached_to_acting_user? ||
+      sys_admin?
+  end
```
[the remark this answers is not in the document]
exbinary: Added a `TODO` in code to be resolved later.

app/controllers/application_controller.rb
```diff
@@ -27,8 +28,12 @@ def default_url_options
 
   private
 
-  def user_not_authorized
-    flash[:alert] = "You are not authorized to perform this action."
-    redirect_to(request.referrer || root_path)
+  def user_not_authenticated(_exception)
```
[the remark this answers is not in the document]
exbinary: Added a `TODO` in code to be resolved later.

app/policies/person_policy.rb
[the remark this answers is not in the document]
exbinary: Added a TODO in code to be resolved later.

exbinary:
Scenarios tested:

* Announcements
  - [x] /announcements is protected
  - [x] /announcements_list is accessible
  - [x] Create new announcement as a guest
  - [x] Approve an annoncement as an admin
* Community Resources
  - [x] /community_resources is protected
  - [x] /community_resources_list is accessible
  - [x] Create new resource as a guest
  - [ ] Cancel from /new fails for a guest -- consistent with main
* Offer|Ask forms
  - [x] Create new ask|offer as a guest
  - [x] /submissions is protected
* Contributions
    * dispatch mode
      - [x] Guest is not able to see /contributions
    * p2p mode
      - [x] Guest is able to see /contributions
      - [x] Able to click 'View' on a contribution
      - [ ] Claim requires login -- consistent with main
      - [ ] Respond button is visible to guests -- consistent with main
* Matches
  - [x] /matches is protected
* User
  - [x] Able to login and logout
  - [x] Sign up is still disabled
  - [x] Admin is able to edit a user account
  - [x] Admin is able to delete a user account
  - [ ] Not able to create a new user -- consistent with main
* People
  - [x] Admin is able to add a person record
  - [x] Admin is able to show and edit a person record
  - [x] Guest is not able to access /people or /people/:id

app/policies/application_policy.rb
```diff
@@ -0,0 +1,67 @@
...
+    # We default all permissions to false, and expect you to override as needed.
+    def resolve
+      original_scope.none
+    end
+  end
+
+  attr_reader :acting_user, :record
+
+  def initialize(acting_user, record)
+    @acting_user = acting_user
```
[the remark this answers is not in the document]
exbinary: Or alternatively, use plain `user` within policies since that seems to be the defacto norm (used in pundit docs).

maebeale (approved):
YESSSSS!!!!!! 

THANK YOU @exbinary @svileshina @armahillo @bhaibel @h-m-m !!!!!

PR #516, opened 2020-06-15
Support deploying to production with docker-compose

SMores:
## This PR
* ~Removes the `@fortawesome/fontawesome-free` npm package. Unfortunately, it wasn't asset-pipeline aware, and so wouldn't work with a precompiled asset pipeline in a production environment~ Moved to #525
* ~Replaces the `@fortawesome/fontawesome-free` npm package with a CDN hosted font-awesome package. I just noticed that this CDN tries to add tracking cookies when you visit a site that uses it, which absolutely sucks, so maybe we should just download these and host them ourselves.~ #525
* Adds a production `Dockerfile`. This is copied from the development one but includes a `rake assets:precompile` step that you probably wouldn't want to use in development. Also reordered some steps to optimize image caching a bit (`bundler install` takes a loooooooong time). I have a hunch that there's some `bundler install --without dev` command or something like that that would speed this up a bit, but my rails is rusty
* Adds a production docker-compose file. It manages the postgres database, the app, and a [caddy server](https://caddyserver.com/) to act as a reverse proxy and static file server (example config in the README)
* Adds a README section on using the production docker-compose set up in production

## Still to come
* Some instructions on updating the app (it's basically just `git pull` and `docker-compose restart`)
* Some shell commands that can automate most of this. I'm imagining basically an `init` that asks the user for a postgres password and a domain name and does all of the setup laid out here, a `start` that runs `docker-compose --build -d`, a `restart` that runs `docker-compose restart` (or maybe `docker-compose down && docker-compose up -d --build`), and an `upgrade` that takes a git ref, checks it out, and then restarts

## Notes
Probably the best/nicest part of this setup is that Caddy runs the ACME challenge on your behalf and obtains Let's Encrypt certs for your domain name, and renews them before they expire.


README.md
```diff
@@ -226,8 +226,66 @@ within it. This allows for any unexpected state data of the user from hanging ar
...
+
+_Note_: These instructions assume you are working with shell access to a system with Docker and docker-compose available (such as a Digital Ocean Docker image).
+
+_Note_: These instructions assume you already have a domain name registered and pointed it at the
+IP address of your server.
+
+1. You'll need to get the code on the server. The recommended way to do this is with a [GitHub Deploy Key](https://developer.github.com/v3/guides/managing-deploy-keys/). Deploy Keys are just SSH keys that are tied to a project, instead of being tied to a user, which means that they can persist independently of the project's membership. You can use [GitHub's guide on generating SSH keys to generate your deploy key](https://help.github.com/en/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent).
+2. After you've generated your deploy key and added it to your project, clone the project.
+3. Create a Caddyfile. It can live anywhere (Mutual Aid Tompkins has ours at `~/.config/Caddyfile`), and it should look like this, replacing `your.domain.name` with your organization's desired domain name:
```
h-m-m, line 240: I'm going to try and run through these now and let you know if I have any questions

h-m-m: The instructions are pretty okay so far. I ran into some problems that are related to my environment because they're happening with other projects, too, now, so I'm going to wipe my setup and try again tomorrow.

As I'm going through this, I think we might need to split up the documentation so it's clearer what steps are for development environments and which are for production.

As we're expecting new people volunteering later this year, I think we'd be best off with development environment guidelines very streamlined and front and center (I imagine people on Windows 10 will get a lot of use out of setting up a Docker development environment). Then instructions that are specific for production could be in a separate file, maybe, with a link?

Definitely open to other ways of organizing this. Let me know what you think

maebeale: I was thinking same thing about the README -- finding parts that can be extracted to other files to make the outline really clear and everything else be links.

.env.ci.example
```diff
@@ -22,3 +22,8 @@ SENDGRID_API_KEY=123
 
 # (optional) override devise hours after which link expires (defaults to 6)
 HOURS_PASSWORD_RESET_LINK_EXPIRES_AFTER=8
+
+# Heroku only! Set to turn off devDependencies pruning
+# See https://devcenter.heroku.com/articles/troubleshooting-node-deploys#missing-modules
+# for more info
+YARN_PRODUCTION=false
```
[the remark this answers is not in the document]
maebeale: Re `.env`, @exbinary is going to work on refactoring those so all possible/default vars are in .env (and then we don't need the ci file anyway, right?). It's my fault for conflating w heroku here.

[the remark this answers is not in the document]
SMores: Yeah that makes sense! We'll need this line regardless of the FA issue, though; it's to prevent heroku from pruning dev dependencies that are needed to compile the client bundles

README.md
SMores: 100% agreed with this! Let me know if you still can't get things working with a fresh environment; I'm _pretty_ sure this is all of the steps I took on a fresh Digital Ocean droplet, but I absolutely might have missed something.

.env.ci.example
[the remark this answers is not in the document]
SMores: This is done!

ansible/files/.env.example
```diff
@@ -22,3 +22,15 @@ export SENDGRID_API_KEY=123
 
 # (optional) override devise hours after which link expires (defaults to 6)
 export HOURS_PASSWORD_RESET_LINK_EXPIRES_AFTER=8
+
```
exbinary, line 25: Just checking that you intended to be changing the `ansible` example file instead of the root `.env` one?

ansible/files/.env.example
```diff
@@ -22,3 +22,15 @@ export SENDGRID_API_KEY=123
...
+# See https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#timeout
+# for more info
+RACK_TIMEOUT_SERVICE_TIMEOUT=20
+
+# Heroku only!
+# ============
+# Set to turn off devDependencies pruning
+# See https://devcenter.heroku.com/articles/troubleshooting-node-deploys#missing-modules
+# for more info (you need to turn this off for Heroku builds to succeed)
+# YARN_PRODUCTION=false
```
exbinary, line 36: I think this is no longer necessary with #560

docker/production/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+
+services:
+  db:
+    image: postgres
+    restart: always
+    volumes:
+      - db:/var/lib/postgresql/data
+    environment:
+      POSTGRES_USER: mat
+      POSTGRES_DB: mutualaidtompkins
```
exbinary, line 16: We probably want installation specific values to not be hardcoded in the main repo, yeah?
Would be fine with extracting them in a later PR though.
There are some further down in this file as well.

README.md
```diff
@@ -220,8 +220,66 @@ within it. This allows for any unexpected state data of the user from hanging ar
...
+   docker-compose run app '/usr/local/bin/rake' secret \
+    | echo "SECRET_KEY_BASE=$(tail -1 -)" > .env
+   ```
+   This will generate a secret token and add it to a file named `.env`, which will be available when
+   `docker-compose` starts.
+6. Generate (using a password manager) a password for the Postgres database, and append it to the `.env` file. Your `.env` file should look something like:
+   ```bash
+   SECRET_KEY_BASE=<A fairly long string of numbers>
+   POSTGRES_PASSWORD=<your postgres password>
+   ```
```
exbinary, line 270: Some thoughts about `.env` files.
1. As of #532 , the main repo `.env` file is no longer empty and has all our default values, so line 262 above might need to append instead of write.
1. Something feels off about having to mess with `.env` files within the container. Keep in mind that [dotenv](https://github.com/bkeepers/dotenv#can-i-use-dotenv-in-production) isn't designed to even load in a typical production setting. Is there an alternate way to set environment variables with docker? That said, i really don't know docker well, so may be completely missing many things!

maebeale:
(@SMores, fyi we merged a conflict in package.json and moved several of the dependences back out of devDependencies)

ansible/files/.env.example
maebeale: @SMores - just wanted to check on this?

Procfile
```diff
@@ -0,0 +1 @@
+web: bundle exec puma -C config/puma.rb
```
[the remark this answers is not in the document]
maebeale: @SMores

PR #525, opened 2020-06-20
Download fontawesome files to serve locally

SMores:
Avoids using a CDN that tries to using tracking cookies, and also lets us use font-url (which the npm package doesn't do, causing it to break in production).


exbinary:
@SMores, can you say more about the `font-url` problem? Is this a known issue with the npm package?

SMores:
Sure, sorry I missed this! When you precompile rails assets, the pipeline renames them, appending a hash code to the filename, as a cache-busting mechanism. The node module isn't specifically built to work with the rails asset pipeline, so the CSS looks for the font files by their statically defined filename, which doesn't include the hash code. The solution to this, in rails land, is the `font-url`, which is rails asset-pipeline-aware, and can update the URL in the CSS at compile-time to have the newly compiled font file names.

PR #542, opened 2020-06-26
Add inexhaustible to listing

maebeale:


app/controllers/contributions_controller.rb
```diff
@@ -58,7 +58,11 @@ def allowed_params
...
 
   def set_contribution
-    type = params["contribution_type"]&.classify || "Ask"
-    @contribution = Listing.where(type: type, id: params[:id]).first
+    type = params["contribution_type"]&.classify
+    if type
+      @contribution = Listing.where(type: type, id: params[:id]).first
+    else
+      @contribution = Listing.find(params[:id])
+    end
```
exbinary, line 66: Any reason not to do `.find(params[:id])` in all cases?

app/models/listing.rb
```diff
@@ -52,7 +53,7 @@ def name
   end
 
   def name_and_match_history
-    "(#{type}-#{all_tags_to_s.upcase})</strong> #{person.match_history}. (#{person.name})"
+    "(#{type}-#{all_tags_to_s.upcase}#{" ***INEXHAUSTIBLE*** " if inexhaustible?}) #{person.match_history}. (#{person.name}#{" --- $" + person.monthly_donation_amount_max.to_s + "/mo left" if person.monthly_donation_amount_max&.to_f > 0.0})"
```
exbinary, line 56: Lot going on on this line :)

maebeale: lol. yup. it's pretty painful. the page this method is called from will be replaced once we add an official "match" page (a version of the Contributions page), so we can get rid of this monstrosity then.

PR #545, opened 2020-06-26
Update listings index

maebeale:
Improve usability Listings index
- Add match_ids scope and apply to instance var
- Allow triage button to accept button_css_class var
- Add Category and change Match Status to link to Match/Matches, if the… …
- Change triage button text
- Add helper to display UrgencyLevel text w html/css
- Needed class_name designation since UrgencyLevel isn't db backed (yet)
- Change category display to match Contributions (in a few places)
- Rearrange columns on Listings index

### DIFF
--- a/app/models/listing.rb
+++ b/app/models/listing.rb
@@ -10,7 +10,7 @@ class Listing < ApplicationRecord
   belongs_to :service_area
   belongs_to :location, optional: true
   belongs_to :submission, optional: true
-  belongs_to :urgency_level, optional: true
+  belongs_to :urgency_level, optional: true, class_name: "UrgencyLevel"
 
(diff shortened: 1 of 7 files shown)


app/models/listing.rb
```diff
@@ -10,7 +10,7 @@ class Listing < ApplicationRecord
   belongs_to :service_area
   belongs_to :location, optional: true
   belongs_to :submission, optional: true
-  belongs_to :urgency_level, optional: true
+  belongs_to :urgency_level, optional: true, class_name: "UrgencyLevel"
```
maebeale, line 13: Bc it's a struct, class_name needed to be listed explicitly

PR #546, opened 2020-06-27
Refactor how Rails handles filter categories passed to and from Vue

h-m-m:
### Why
The contribution filtering works nicely, but difficult to understand. We'd like to add new filters, and that's hard at the moment. I wrote the original code and am unhappy with how confusing it is. Its confusing complexity is what makes adding or editing filters more difficult for us, and also slows down bringing people into the project and up to speed on how all this stuff works. 

### What

The big change is making each type of filter its own class.
Each class
- accepts an ActiveRecord relation
- knows how to filter that relation by request parameters passed from the controller

Also used this opportunity to:
- make the controller code easier to follow
- simplify the logic around conditionally passing links to the frontend (do we show the link to contact someone? does the user have permission to see this?
- work on our nomenclature for clarity & simplicity
 
And hopefully this will make adding new filters easier in the future

### Testing

Updated some tests. Could probably use more work in this regard

### Next Steps

Some next steps are already called out in the comments when this PR was in Draft status.
Also: add a filter for Date now that it should be easier to do so

### Accessibility

We were very careful to leave the existing UI unchanged in any visible way, so this should be accessibility-neutral

### Security

Hopefully this improves the security what with less dynamic/functional code and code that's easier for future devs to maintain

### Pre-Merge Checklist


app/filters/contribution_type_filter.rb
```diff
@@ -0,0 +1,22 @@
...
+  attr_reader :parameters
+
+  def initialize(params)
+    @parameters = params
+  end
+
+  def scopes
+    classes = parameters.blank? ? ALL_ALLOWED_TYPES : parameters.keys
+    classes.intersection(ALL_ALLOWED_TYPES).map do |type|
+      type.constantize.matchable
```
h-m-m, line 19: Is this right — that `matchable` is what we want here?

h-m-m (commented):
I added a bunch of comments to stuff because it helps me.

Sorry, @maebeale, for the flipping back and forth about what goes in instances and what goes in classes. I added on to your work with what I thought I remembered @exbinary was interested in, but I might have misremembered

app/filters/contact_method_filter.rb
```diff
@@ -0,0 +1,13 @@
...
+  def self.options
+    {
+      name: 'Contact Methods',
+      filters: FilterOptionsBlueprint.render_as_hash(ContactMethod.enabled.distinct(:name))
+    }
+  end
+
+  def filter(scope)
+    return super unless parameters
+    scope.joins(:person).where(people: { preferred_contact_method: parameters.keys })
```
exbinary, line 11: I'm noticing all the filters use `parameters.keys`. Could we get the keys in `BrowseFilter` or `BaseFilter.new` and make just the keys available to individual filters?

Relatedly, the shape of `parameters` is not clear from looking at the code. Could be solved by looking at specs, or maybe a comment describing how params will come in?

app/filters/category_filter.rb
```diff
@@ -0,0 +1,17 @@
+class CategoryFilter < BasicFilter
+  def self.options
+    {
+      name: "Categories",
+      # Currently only filtering by top-level categories
+      filters: FilterOptionsBlueprint.render_as_hash(Category.roots)
+    }
```
exbinary, line 7: Serializing inside the filters doesn't feel quite right. Is it possible to have `options` return a relation as the value of `filters` and let the controller handle serialization?
```suggestion
    {
      name: "Categories",
      # Currently only filtering by top-level categories
      filters: Category.roots
    }
```

This might require introducing another blueprint at the level of a filter that then has an `association :filters`.

exbinary (approved):
Looks great, @h-m-m ! A few comments inline but overall really like how this is shaping.

app/filters/category_filter.rb
h-m-m: The only with your suggestion is that the Ask/Offer distinction is one I'm doing manually:

https://github.com/rubyforgood/mutual-aid/blob/39a0d099aaa860edf396a000e86ee8bdb8397237/app/filters/contribution_type_filter.rb#L1-L7

app/filters/contact_method_filter.rb
h-m-m: > the shape of parameters is not clear from looking at the code

This is really the key point here. Thanks for surfacing this, because this point is helping me find the most confusing points.

I started writing some tests around this and got stuck because it was *hard*. I think I've introduced a small, modest mess in `BrowseFilter#contributions` and cleaning up the parameters structure here is dependent on cleaning that up, too.

There's a part of me that wants to get this PR out first and then focus on cleaning up that problem. I want to do that because this PR has taken a very long time and not directly for any code-specific reasons. This PR seems so big to me already, and I'd really like to focus on the refactor I'll need here as a separate thing from the other changes in this PR.

I'm open to suggestions. Let me know what you think

app/filters/contribution_type_filter.rb
maebeale: yes. we'll need to add that scope to `CommunityResource` to return objects that are still published

app/controllers/contributions_controller.rb
```diff
@@ -46,21 +56,11 @@ def peer_to_peer_mode?
...
-
-    allowed_params.to_h.filter { |key, _v| BrowseFilter::ALLOWED_PARAMS.keys.include? key}.tap do |hash|
-      hash.each_key { |key| hash[key] = hash[key].keys}
-    end
+    { show_view_path: true }
   end
 
   def allowed_params
-    @allowed_params ||= params.permit(:format, **BrowseFilter::ALLOWED_PARAMS)
+    params.permit(:format, BrowseFilter::ALLOWED_PARAMS_FILTER)
```
h-m-m, line 63: TODO: rename `ALLOWED_PARAMS_FILTER`

app/models/browse_filter.rb
```diff
@@ -1,48 +1,58 @@
...
-    parameters.keys.reduce(model.matchable) do |scope, key|
-      filter = FILTERS.fetch(key, ->(_condition, s) {s})
-      filter.call(parameters[key], scope)
+    # parameters.keys.reduce(model.unmatched) do |scope, key|
+    #   filter = FILTERS.fetch(key, ->(_condition, s) {s})
+    #   filter.call(parameters[key], scope)
+    # end
+
+    FILTERS.reduce(model.unmatched) do |relation, filter|
+      filter.filter(relation, parameters)
```
h-m-m, line 55: TODO: this can probably be safely deleted

app/controllers/contributions_controller.rb
```diff
@@ -3,11 +3,21 @@
...
+  #   A higher-level category or other grouping of filter options. Example: Contact Method can be
+  # a *filter group* that then has *filter options* for things like "email" or "text message"
+  # Filter Option —
+  #   An individual item that can be chosen to change what's filtered. Each *filter option* is 
+  # associated to one and only one *filter group*
+
   def index
-    @filter_types = FilterTypeBlueprint.render([ContributionType, Category, ServiceArea, UrgencyLevel, ContactMethod])
-    filter = BrowseFilter.new(filter_params)
+    @filter_groups = BrowseFilter.filter_groups_json
```
h-m-m, line 18: TODO: rename `filter_groups` -> `filter_groupings`

app/controllers/contributions_controller.rb
h-m-m: @exbinary suggested that calling it `ALLOWED_PARAMS` is probably sufficient

app/filters/category_filter.rb
exbinary: Ah yes, i see the challenge. I had a go at extracting the serialization out in 8b8c8a94b5dd1e50f256e12abbd3071550ebed61, curious your thoughts?
In any case, i don't think this question should hold up this PR.

exbinary (commented):
@h-m-m , i merged main, resolved conflicts and pushed to this branch.
Should be good to merge once you give it a quick look.

app/filters/contribution_type_filter.rb
```diff
@@ -0,0 +1,22 @@
...
+      { id: 'ContributionType[Offer]', name: 'Offer' }
+    ]}
+  end
+  ALL_ALLOWED_TYPES = ['Ask', 'Offer'].freeze
+
+  attr_reader :parameters
+
+  def initialize(params)
+    @parameters = params
+  end
```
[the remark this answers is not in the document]
h-m-m: Note: discussed this with @exbinary in our regular meeting and added a commit per our discussion

h-m-m:
Notes on what we're thinking of following up on as a subsequent refactoring:

- [ ] Removing `paramters.keys` references in the filters so filter classes can scan the full parameter list as they see fit ([comment](https://github.com/rubyforgood/mutual-aid/pull/546#discussion_r611060827)
- [ ] Move serialization concerns out of the filter classes, perhaps by breaking up `filter_grouping` in to one method that returns the filter grouping's name and another method that returns the filter options for that filter grouping ([comment](https://github.com/rubyforgood/mutual-aid/pull/546#discussion_r611322656))

Issue #479, opened 2020-06-09
Heroku isn't finding the font-awesome icons


maebeale:
for some reason heroku isn't rendering any of the font-awesome icons

<img width="1042" alt="Screen Shot 2020-06-11 at 8 22 32 PM" src="https://user-images.githubusercontent.com/7607813/84451538-5351f600-ac21-11ea-8f51-92c276730006.png">

<img width="699" alt="Screen Shot 2020-06-11 at 8 21 50 PM" src="https://user-images.githubusercontent.com/7607813/84451540-55b45000-ac21-11ea-8305-1daf376214c5.png">

PR #560, opened 2020-07-01
Serve stylesheets with webpack

exbinary:
Switch css delivery from sprockets to webpack.
Fixes the font-awesome issue, #479.

This is the fourth direction we've explored to fix the font-awesome woes:
1. The current _broken_ approach which pulls in the `npm` package but serves it from sprockets.
1. #525, which downloads font-awesome assets into our repo, serving them via sprockets.
1. The [`font-awesome-fix`](https://github.com/maebeale/mutual-aid/compare/font-awesome-fix) branch, which switches to a gem served via sprockets.
1. Now this :)

The advantages of this approach seem to be:
* The npm package is kept more up to date than the gem.
* There's a simple upgrade path in comparison to downloading assets into our repo.
* Gets us pretty close to being able to completely dump sprockets, letting webpack deliver everything (only images remain). Having one asset delivery mechanism feels less confusing to me, but i recognize it may be debatable whether this is an advantage!

Deployed to the [test app](https://mutual-aid-app-test.herokuapp.com/) and all seems to be working there.

Definitely open to discussion.

/cc @SMores, @h-m-m


app/javascript/pages/Ask.vue
```diff
@@ -96,7 +96,7 @@ export default {
 }
 </script>
 
-<style>
+<style scoped>
```
maebeale, line 99: what does this do, keep this style only available to this page?

exbinary: It scopes the styles just to this component and any child components.

PR #566, opened 2020-07-03
Serve images with webpack and remove Sprockets

exbinary:
Completes the switch to webpack by switching image delivery to webpack and removing sprockets altogether...

... with much gratitude for your many years of solid service, dear old friend 🙏🏾❤️!

Notes:
* In rails views, use `image_pack_tag` instead of `image_tag`.
* The `assets:precompile` rake task is now an alias for `webpacker:precompile`
* The `assets:clobber|clean` rake tasks are now gone, use the corresponding webpacker ones instead. Note that `heroku` tries to run `:clean` late in its deployment; this fails but heroku logs the error and keeps pushing, so i don't think we need to change anything.


app/views/devise/registrations/new.html.erb
```diff
@@ -1,7 +1,6 @@
 <div class="row text-center login-screen" style="vertical-align:middle;">
   <div class="col-md-6 col-md-offset-2 logo-image text-center">
-    <%= image_tag @current_organization&.logo_url.present? ? @current_organization&.logo_url : "logo.png",
-                  width: 300, alt: 'mutual-aid-app logo'%>
+    <%= render 'shared/organization_logo' %>
```
maebeale, line 3: why would this go in /shared and not under layouts?

exbinary:
Thanks for the super quick review @maebeale! I added a tangential [commit](https://github.com/maebeale/mutual-aid/pull/566/commits/e41f7b75b3164f5974b7a829fc93aadeaad68f7d) after you reviewed. Want to make sure you see it because i'm proposing a minor change to the changelog format we've been using.

exbinary:
> Thanks for the super quick review @maebeale! I added a tangential [commit](https://github.com/maebeale/mutual-aid/pull/566/commits/e41f7b75b3164f5974b7a829fc93aadeaad68f7d) after you reviewed. Want to make sure you see it because i'm proposing a minor change to the changelog format we've been using.

Actually, NM -- pulled it out into separate PR :).

app/views/devise/registrations/new.html.erb
exbinary: I was going to ask about this also, but from the other side :).

The convention (?) i've seen is that partials shared between multiple pages go under `view/shared`. I haven't actually seen partials created under `layouts/` unless they are specifically only for use by top level layouts.

PR #570, opened 2020-07-09
Show nested categories differently in categories index

maebeale:
- On categories index, nest child categories below root (parent) categories and display rows in greyscale

After:
Screen Shot 2020-07-09 at 2 30 58 PM

Before:
Screen Shot 2020-07-09 at 2 29 32 PM

### DIFF
--- a/app/assets/stylesheets/categories.scss
+++ b/app/assets/stylesheets/categories.scss
@@ -1,0 +1,4 @@
+.grey-row {
+  color: #7a7a7a;
+  background-color: #e3e3e3;
+}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -4,6 +4,7 @@
 @import './custom_spacing_helpers';
 @import './bulma_simple_form_customization';
 
+@import './categories';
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/categories.scss
```diff
@@ -0,0 +1,4 @@
+.grey-row {
+  color: #7a7a7a;
+  background-color: #e3e3e3;
+}
```
exbinary, line 4: FYI and future reference, bulma also provides these:
https://bulma.io/documentation/helpers/color-helpers/

maebeale: thanks! they're a bit more heavy-handed (or colors) than we were going for. i actually want to lighten that grey even more now that i'm looking at it again! :)

exbinary: Hmm, a bit surprised their range doesn't cover it. IIRC, they had like a dozen shades of gray!

maebeale: duh. you're right. they can be mixed and matched!

PR #574, opened 2020-07-09
Replace custom css w bulma classes on categories index

maebeale:
- Remove extra scss file and custom class
- Add bulma classes to style the subcategory rows on Categories index

Screen Shot 2020-07-09 at 5 34 41 PM


maebeale:
thank you, @exbinary !!!

PR #581, opened 2020-07-22
Get logo_url from current_organization

maebeale:
@current_organization wasn't defined, so the login logo was defaulting to the stock image


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+### Enhancements
```
exbinary, line 4: Oh, is this more of a bug?

PR #591, opened 2020-07-23
Implement in-memory ActiveJob to background response import

### LINKED ISSUE #585: Add ActiveJob/Delayed Job to process importers in the background

exbinary:
**Important**: this implementation has two major limitations:
1. Uses the `:async` queue backend which is in-memory only and isn't usually recommended for production. For now, to support imports only which typically happen before an installation is live, it might suffice. It does have the distinct advantage of not needing any additional infrastructure
2. Passes the _whole_ uploaded file contents into the background job as a string. This will obviously not scale and we'll have to monitor how feasible this is.

Resolves #585.

### DIFF
--- a/app/controllers/submission_response_imports_controller.rb
+++ b/app/controllers/submission_response_imports_controller.rb
@@ -5,8 +5,14 @@ class SubmissionResponseImportsController < ApplicationController
 
   def create
     uploaded_file = params[:submission_response_import]
-    importer = Importers::SubmissionResponseImporter.new(current_user, uploaded_file.original_filename)
-    importer.import(uploaded_file.path)
+
+    SubmissionResponseImportJob.perform_later(
+      user_id: current_user.id,
+      file_name: uploaded_file.original_filename,
+      file_contents: uploaded_file.read,
+    )
+
+    flash[:notice] = 'Your file has been uploaded and is being imported'
     redirect_to landing_page_admin_path
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/importers/base_importer.rb
```diff
@@ -109,7 +109,18 @@ def import(path)
...
+    import_rows(rows)
+  end
+
+  def import_string(string)
+    Rails.logger.info("START IMPORT------------#{Time.now}")
+    records_report("initial")
+
+    rows = CSV.parse(string, headers: true)
+    import_rows(rows)
+  end
```
exbinary, line 121: Not sure if you want different logging in these two alternate methods. Just kept it consistent with how it currently works.

Ideally, we could rename `import` as well, but i didn't want to break the existing interface.

maebeale: love it!!! THANK YOU!!!!!

app/controllers/submission_response_imports_controller.rb
```diff
@@ -5,8 +5,14 @@ def new
 
   def create
     uploaded_file = params[:submission_response_import]
-    importer = Importers::SubmissionResponseImporter.new(current_user, uploaded_file.original_filename)
-    importer.import(uploaded_file.path)
+
+    SubmissionResponseImportJob.perform_later(
+      user_id: current_user.id,
+      file_name: uploaded_file.original_filename,
```
maebeale, line 11: this now gets imported as "csv" -- idk why it lost the original filename

maebeale: but, LGTM for now!!!

PR #594, opened 2020-07-24
Only show visible subcategories on forms

maebeale:
- All subcategories for any visible Category were displaying, but the visible scope needs to be applied to subcategories as well

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -10,6 +10,12 @@ class Category < ApplicationRecord
     inverse_of: :parent
   )
 
+  has_many(:visible_subcategories, -> { visible.order(:display_order, :name) },
+           class_name: "Category",
+           foreign_key: :parent_id,
+           inverse_of: :parent
+  )
+
   validates :name, presence: true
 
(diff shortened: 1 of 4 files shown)


app/models/category.rb
```diff
@@ -10,6 +10,12 @@ class Category < ApplicationRecord
     inverse_of: :parent
   )
 
+  has_many(:visible_subcategories, -> { visible.order(:display_order, :name) },
+           class_name: "Category",
+           foreign_key: :parent_id,
+           inverse_of: :parent
+  )
+
```
exbinary, line 18: I'm curious why this needs to be a separate `has_many` instead of a `scope`?

maebeale:
[image: Screen Shot 2020-07-24 at 10.20.47 AM.png]

app/models/category.rb
maebeale: @exbinary I thought it needed to be an assn, per this language in the blueprint:

Screen Shot 2020-07-24 at 10 20 47 AM

PR #599, opened 2020-07-25
Fix pending specs for SubmissionForm updating

exbinary:
Required looking up existing submission_responses by custom_form_question_id.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,2 @@
 --require spec_helper
+--no-profile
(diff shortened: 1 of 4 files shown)


.rspec
```diff
@@ -1 +1,2 @@
 --require spec_helper
+--no-profile
```
exbinary, line 2: Better default so that spec results are visible without having to scroll up. Can be overridden by passing `--profile N` to rspec.

PR #624, opened 2020-08-11
Migrate multi listing submissions

exbinary:
This got complicated when we started to consider sub-categories!

See af063eca12d82ded8a2568d1c61a54320b7b60e2 for my attempt to deal with them. I'm not sure how well the code speaks for itself; happy to get on a call if that would be easier/better.

The `one_time:split_listings` rake task _should_ be idempotent on repeated runs, though it will always have to pull and examine any listings with sub-categories (since the query looks for any listing with multiple tags). I formatted the `puts` output to be csv compatible with 3 columns:
1. an indicator for visual scanning (`#` - listing being processed, `-` listing after processing, `+` newly created listings),
1. the listing id (original or new),
1. tag_list

Will definitely need some 👀 to verify my specs/logic and do some more testing.

### DIFF
--- a/lib/tasks/one_time.rake
+++ b/lib/tasks/one_time.rake
@@ -3,16 +3,17 @@ require_relative './support/listing_converter'
 namespace :one_time do
   desc "split listings with multiple tags into one listing for each tag"
   task split_listings: :environment do
+    converter = ListingConverter.new
+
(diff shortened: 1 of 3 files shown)


exbinary:
Force pushed with a simple TagSplitter algorithm, see af063eca12d82ded8a2568d1c61a54320b7b60e2.

maebeale (approved):

PR #673, opened 2020-09-04
Don't rely on fixed ids in contributions_spec

exbinary:
Ran into a case where the id conflicted with an existing record in the test db (possibly from a seed).


h-m-m (approved):
Thanks for this, Lasitha. @viamin had this exact same problem.

viamin:
I did find that running the setup code exactly like circleci did seem to avoid the spec failure.

exbinary:
> I did find that running the setup code exactly like circleci did seem to avoid the spec failure.

Good to know!

PR #676, opened 2020-09-04
Fix devise emails

maebeale:
We were getting a SendGrid error when using SignUp bc the emails weren't sending, prompting us to clean up ENV vars.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -18,7 +18,7 @@ Devise.setup do |config|
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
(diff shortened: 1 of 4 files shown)


config/initializers/devise.rb
```diff
@@ -18,7 +18,7 @@
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  config.mailer_sender = "#{ENV["SYSTEM_EMAIL"]}"
+  config.mailer_sender = "#{ENV["SMTP_USERNAME"]}"
```
exbinary, line 21: I think this should also be `SMTP_FROM_EMAIL`? Can consider/fix later since i think the one in our mailers is overriding it (according to the comment above this line).

exbinary (approved):
woohoo!

Issue #663, opened 2020-09-03
Relax sanitization rules so users can do basic things


maebeale:
Currently users can't use bold, divs with classes, etc

An example in prod: 
http://mutual-aid-ggf.herokuapp.com/about

* They want to use bold, headings, ol/ul
* Here's their html
```
<div class="title is-3">About us</div>
<ul>
  <ol>
    <li>
      <span class='title is-5'>What is Mutual Aid?</span>
      <br><br>
      <span class="has-text-weight-bold">Mutual Aid is</span> neighbors supporting neighbors in the spirit of solidarity to meet survival needs, as opposed to non-profit/charity or government work where people are dependent on a central organization. We recognize that <span class="has-text-weight-bold">the community knows best</span> what its needs are and how to meet them.
      <br><br>
      <span class="has-text-weight-bold">Mutual Aid</span> supports all our neighbors, but we recognize current and historical oppression and want to prioritize meeting the needs of our most vulnerable neighbors including <span class="has-text-weight-bold has-text-italic">the poor, sick, elderly, disabled, undocumented, LGBTQ+, Black, Brown, Indigenous and/or people of color.</span>
      <br><br>
      <span class="has-text-weight-bold">Mutual Aid</span> is political and anti-oppressive by nature, and focused on creating <span class="has-text-weight-bold has-text-italic">long-term, sustainable systems of interdependent community care</span> that ultimately lead to a more resilient and self-reliant region. We want to foster a culture of collectivism in the Greater Glens Falls area that operates outside individualism and existing government and non-profit systems.
      <br><br><br><br>
    </li>

    <li>
      <span class='title is-5'>Why is Mutual Aid necessary?</span>
      <br><br>
      <span class="has-text-weight-bold">Mutual Aid</span> fills gaps. It is homegrown and grassroots. It protects the privacy of people who are asking and people who are giving. No strings are attached, in order to avoid power dynamics. It levels the playing field and honors all people. It connects people to existing resources and meets a need for those who fall through the cracks. It supports and strengthens existing organizations.
      <br><br>It is central, human-focused, and, at its roots, it is <span class="has-text-weight-bold">ANTI-OPPRESSION</span>.
<br><br><br><br>
    </li>

    <li>
      <span class='title is-5'>What are the foundations of Mutual Aid?</span>
      <br><br>
      <ul>
        <li>Cooperation, not competition</li>
        <li>Non-hierarchical, non-bureaucratic </li>
        <li>Egalitarian</li>
[...]

PR #683, opened 2020-09-05
Add more tags to the HtmlSanitizer allow list

viamin:
Add b, h{1,2,3,4,5,6}, and span to allowed tags in the HTMLSanitizer model
Also add a spec file for this (not all methods and allowed attributes are tested)

Addresses https://github.com/rubyforgood/mutual-aid/issues/663


exbinary (approved):
Looks great! Appreciate the solid spec as well!

PR #685, opened 2020-09-06
Fix docker buildfile

cflipse:
Was resulting in an error when attempting to run the build:
  app_1    | /usr/local/bundle/gems/thor-1.0.1/lib/thor/parser/arguments.rb:133:in `parse_numeric': Expected numeric value for '--port'; got "${APP_PORT}" (Thor::MalformattedArgumentError)

There are two things that aren't quite set up right here:

The build arg was not in scope -- used args need to be _below_ the
`FROM` statement to be accessible.

CMD appears to not allow variable expansion, because it is processed at
runtime rather than build time.  Instead, we can export the `PORT`
envar, which will be picked up by `config/puma.rb` when starting rails.

### DIFF
--- a/docker/development/app/Dockerfile
+++ b/docker/development/app/Dockerfile
@@ -1,10 +1,9 @@
-ARG APP_PORT
-
 FROM ruby:2.7.0-alpine
+
+ARG APP_PORT=3000
 
 RUN bundle config --global frozen 1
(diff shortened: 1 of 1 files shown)


grenewode (approved):
This all looks good! We should think about if there's a way to run ci tests on these so we avoid these kinds of breakages in the future!

PR #686, opened 2020-09-06
Install the annotate gem rake file and run it

viamin:
The Gemfile contains the `annotate` gem, but it looks like it was never run. This PR is the result of running `rails g annotate:install` and `rails annotate_models` with the default settings. 

The `annotate` gem adds comments to the top of every ActiveRecord model class with the column names, indexes, and foreign keys. It makes it easy to see what columns (and hence auto-generated attributes) are available on a class without having to skim through `db/schema.rb`.


exbinary (commented):
This is a neat tool! Thanks for configuring and running it.

Does it automatically generate new annotations when schema.rb changes?

TBH, i'm not sure i'd prefer having these annotations...

I don't think i need to look up this info often enough to warrant the extra comments, which come with a cost: having to scroll past them. While this is small cost, it might add up since its incured every time i open a model file.

And when i do need this info, my vim-ruby plugin gives me a handy shortcut to jump from a model straight to the corresponding definition in schema.rb (and back again), which i've found is good enough.

That's of course my personal view and i'd love to hear how others feel.

If we do go forward with it i might suggest:
* moving annotations to the bottom of source files
* only adding them to model classes and not specs (i could go either way on factories).

bhaibel:
I've personally found this a useful tool when navigating codebases that I don't know well, or codebases that are too large for anyone to know well. I think in an open-source project context, this could make the project more welcoming for new contributors.

`annotate` does auto-update annotations when you run migrations, so these comments do stay up-to-date.

I'm comfortable with the suggestion of moving them to the bottom, but I would like to offer a counterpoint -- if we start using something like yardoc for internal documentation, keeping them at the top might integrate nicely with that. (I stress the _might_. I'm familiar with both `yardoc` and `annotate`, but not on the same project, so I don't actually know if they play nicely together.)

I've never seen schema annotations on factories before, but they feel clarifying to me there.

I agree that they feel extraneous & verbose in the context of spec files.

viamin:
I've moved the annotations to the bottom of classes and factories, and removed them from specs (and updated the config to automatically add new annotations this way.)

viamin:
It's pretty easy to move it back to the top if we change our minds about it later.

exbinary (approved):
Looks great! Thanks for initiating this @viamin and your thoughts @bhaibel. With the adjustments i feel pretty good about it and am seeing how it could be very helpful. 🙏🏾🔍

PR #691, opened 2020-09-10
Option to display navbar

svileshina:
The issue is attached - there needs to be an option as to whether the navbar should show up on the Community Resources and the Annoucements pages. So I added an option for that in system settings that can be toggled. 

Question: 
the lists /annoucements-list and /community-resources-list are both handled by PublicPagesController, so I added the same method in there as well but I am not sure if that was intended or not (last commit)

Also should index also have the display_navbar switch or should it always be without_navbar?

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -3,7 +3,7 @@ class AnnouncementsController < ApplicationController
   before_action :authenticate_user!, except: [:new, :create]
   before_action :set_announcement, only: [:show, :edit, :update, :destroy]
 
-  layout "without_navbar", only: [:new, :create, :show]
+  layout :determine_layout
 
   def index
     @announcements = Announcement.order(created_at: :desc)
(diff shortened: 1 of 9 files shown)


app/controllers/announcements_controller.rb
```diff
@@ -3,7 +3,7 @@ class AnnouncementsController < ApplicationController
   before_action :authenticate_user!, except: [:new, :create]
   before_action :set_announcement, only: [:show, :edit, :update, :destroy]
 
-  layout "without_navbar", only: [:new, :create, :show]
+  layout :determine_layout, only: [:new, :create, :show]
```
exbinary, line 6: Minor: i _think_ we can drop `:create` from this list since there isn't really a view associated with that action.

app/controllers/announcements_controller.rb
```diff
@@ -47,6 +47,10 @@ def set_announcement
       @announcement = Announcement.find(params[:id])
     end
 
+    def determine_layout
+      "without_navbar" unless @system_setting.display_navbar
```
exbinary, line 51: Minor: i think rails generates a `display_navbar?` method with the `?` suffix. Might a tad more idiomatic/expressive to use that one?

exbinary (approved):
Looks great! 🚀 A couple of very minor comments.

Issue #626, opened 2020-08-27
Implement redesigned Vue landing page


maebeale:
![Screenshot_20200701-203256](https://user-images.githubusercontent.com/7607813/91501270-45605880-e893-11ea-8e7f-fcdc3b046700.jpg)
![Lama web design 1](https://user-images.githubusercontent.com/7607813/91501272-45f8ef00-e893-11ea-876b-76fcd5f27722.jpg)
![Lama design notes 2](https://user-images.githubusercontent.com/7607813/91501274-485b4900-e893-11ea-8c0e-c6c773683386.jpg)

slaloggia:
Giving this a shot!

svileshina:
Excited for the weekend - let us know if you have any questions!

PR #699, opened 2020-09-12
626 landing page redesign

slaloggia:
Implements redesigned wireframe in issue #626 

1. Restructures existing components
2. Creates Carousel component for Announcements
3. Announcement title conditionally renders as link to full announcement 
4. View links in action boxes go to filtered Contributions (page is authenticated, redirects to Login if no current user)
5. Renders title with name of organization

### DIFF
--- a/app/controllers/public_pages_controller.rb
+++ b/app/controllers/public_pages_controller.rb
@@ -9,6 +9,11 @@ class PublicPagesController < PublicController
     @announcements = Announcement.where(is_approved: true).published
   end
 
+  def announcements_json
+    announcements = Announcement.where(is_approved: true).published
+    render json: announcements
+  end
+
   def community_resources
     @admin_status = params[:admin] ? YAML.load(params[:admin]) : current_user&.admin_role?
     @community_resources = CommunityResource.where(is_approved: true).published
--- a/app/javascript/pages/Orientation.vue
+++ b/app/javascript/pages/Orientation.vue
@@ -16,7 +16,7 @@
               <a href="/asks/new" class="button is-primary">Submit an Ask</a>
             </div>
             <div>
-              <a href="/contributions">View asks</a>
+              <a href="/contributions?ContributionType[Ask]=1">View asks</a>
             </div>
           </div>
         </div>
@@ -27,7 +27,7 @@
               <a href="/offers/new" class="button is-primary">Submit an Offer</a>
             </div>
             <div>
-              <a href="/contributions">View offers</a>
+              <a href="/contributions?ContributionType[Offer]=1">View offers</a>
             </div>
           </div>
         </div>
--- a/app/javascript/pages/Orientation.vue
+++ b/app/javascript/pages/Orientation.vue
@@ -1,50 +1,60 @@
 <template>
   <div class='orientation section'>
-    <div class='columns actions is-centered'>
-      <div class='column is-one-third'>
-        <div class='box action'>
-          <h2 class='subtitle is-3'>Need some help?</h2>
-          <div class='action-buttons'>
-            <a href="/asks/new" class="button is-primary">Submit an Ask</a>
+      <div class='columns actions is-centered'>
+        <div class='column is-one-third'>
+          <div class='box action'>
+            <h2 class='subtitle is-3'>Need some help?</h2>
+            <div class='action-buttons'>
+              <a href="/asks/new" class="button is-primary">Submit an Ask</a>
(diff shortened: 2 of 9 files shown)


h-m-m:
Thanks for all this work, @slaloggia! 

I remember you showing a demo of it in the meeting yesterday. Do you mind adding some screenshots of that to this PR?

slaloggia:
image

slaloggia:
image

app/controllers/public_pages_controller.rb
```diff
@@ -9,6 +9,11 @@ def announcements
     @announcements = Announcement.where(is_approved: true).published
   end
 
+  def announcements_json
+    announcements = Announcement.where(is_approved: true).published
+    render json: announcements
+  end
```
h-m-m, line 15: To be more idiomatically Rails-y, you could roll this into the `announcements` method above. Here's a similar example elsewhere in this project: https://github.com/rubyforgood/mutual-aid/blob/9373a91680ab76603e706f4ea2eb22f3fe8cf936/app/controllers/contributions_controller.rb#L11-L14

And then Rails should pick the correct option if you explicitly tell it `<path>.html` versus `<path>.json` in your URL

app/javascript/pages/Orientation.vue
```diff
@@ -1,81 +1,125 @@
...
+          <AboutSection id='landing_page_text_what' title='What is Mutual Aid?' :description="landing_page_text_what"/>
+        </div>
+        <div class='column'>
+          <AboutSection id='landing_page_text_who' title='Who are we?' :description="landing_page_text_who"/>
+        </div>
+        <div class='column'>
+          <AboutSection id='landing_page_text_how' title="Where do you fit in?" :description="landing_page_text_how"/>
+        </div>
+      </div>
+      <hr>
```
h-m-m, line 51: I really like how this section, with the `hr`s and the columns, comes out looking

h-m-m (approved):
I feel this has so many nice improvements to the page

app/controllers/public_pages_controller.rb
slaloggia: I had a feeling that's what was happening there, but I wasn't sure. Thanks!

slaloggia: Would you like me to make that change before the merge, or should that be a separate pr?

exbinary: If you're ok making the change, would be best to do that before merging, IMO. Less to keep track of :)

app/javascript/pages/orientation/Carousel.vue
```diff
@@ -0,0 +1,77 @@
+<template>
```
exbinary, line 1: Minor: looks like a lot of this file got indented with 4 spaces instead of 2? Would love to get a linter implemented so this kind of thing can be spotted automatically 🔍.

app/javascript/pages/orientation/Carousel.vue
```diff
@@ -0,0 +1,77 @@
...
+
+export default {
+    data() {
+        return {
+            auto: true,
+            arrow: true,
+            pauseHover: false,
+            arrowBoth: true,
+            arrowHover: true,
+            interval: 5000,
```
exbinary, line 37: Minor: `data` is typically used to setup up 'observed' variables, so that whenever one changes, vue knows to track anywhere it's used and propagate the change. IIUC, the values above are static and so don't technically need to be in `data`. A couple of alternatives are to hardcode them directly into the view (probably the simplest solution) or, if you want to make it more apparent what the available options are, define them as [`$options`](https://vuejs.org/v2/api/#vm-options).

app/javascript/pages/orientation/Carousel.vue
```diff
@@ -0,0 +1,77 @@
...
+                    </div>
+                </section>
+            </b-carousel-item>
+        </b-carousel>
+    </section>
+</template>
+
+<script>
+import { Carousel } from 'buefy';
+const fetch = require("node-fetch");
```
exbinary, line 27: Minor: for consistency, probably better to use `import` here instead of `require`.

exbinary (commented):
Apologies for the late review! This looks great, just a couple minor comments.
Thank you!

exbinary:
Hehe, looks like this got merged in the meantime @slaloggia . I've created a separate issue to track the few outstanding things: #704 .

PR #700, opened 2020-09-13
Update docker config

cflipse:
Update the docker config.

Attempt to streamline the development config, and the test-run; there are a few things that we can do to clear up the structure going forward, but ... at least this should work


docker/development/docker-compose.yml
```diff
@@ -38,23 +38,39 @@ services:
...
+      - gem_cache:/usr/local/bundle/gems
+      - node_modules:/usr/src/app/node_modules
     depends_on:
       - db
       - email
     environment:
       <<: *x-shared-postgres-environment
       <<: *x-email-defaults
-      RAILS_ENV: docker-development
+      RAILS_ENV: development
```
exbinary, line 55: Such a small diff causing much loss of hair!

exbinary (approved):
Awesome, thank you for digging into this!!

PR #703, opened 2020-09-17
Add optional asdf-vm support and development Makefile

thestephenmarshall:
## asdf-vm Version Manager

When it comes to managing various versions of your tools like ruby and node, [asdf-vm](https://asdf-vm.com/#/) is an excellent version manager, attempting to cover all needs. I added the very basic `.tool-versions` file based on the `mutual-aid` specified versions of `ruby`, `yarn` and `nodejs`. If a contributor is using `asdf`, they can now make use of their same familiar version manager! 🥳 

## Docker Startup

Looking through the repo, I found that startup might be enhanced a little by adding a `Makefile` so that now starting the app in Docker is just `make start` (etc.).  This could most definitely be expanded upon as development continues.


docker/development/Makefile
```diff
@@ -0,0 +1,31 @@
+SHELL=/bin/bash -o pipefail
+
+build:
+	docker-compose up --build
+
+rspec:
+	docker-compose run app bin/rspec
+
+seed:
+	docker-compose run -e SYSTEM_EMAIL=${EMAIL}  -e SYSTEM_PASSWORD=${PASSWORD} app rails db:prepare db:seed
```
thestephenmarshall, line 10: Example use: `EMAIL=hello@world.com PASSWORD=abc123 make seed`

exbinary (approved):
Sweet, thanks! 🙌🏾

PR #705, opened 2020-09-21
Only start development services by default via make

indiebrain:
There's no need to run the test service when booting the application
for development via docker and Make


exbinary (approved):

indiebrain:
Thanks, mate!

Issue #707, opened 2020-09-21
Implement Map on Contributions Page


kevinschlabach:
**Story**
As a community member, I want to see a map view on the contributions page, so that I have a place to eventually see contributions visualized geographically (future stories)

**How does it work now**
contributions pages can display contributions by tile or list view only

**How should it work**
A map view is presented as a 3rd option, and a base map is displayed on the page
This story focuses on showing a map as a first pass, future story #708 will render plot points onto it

==========
**Notes**

_Workflow_
- logged in user navigates to http://mutual-aid-demo.herokuapp.com/contributions
- toggles to "Map View"
- right hand area shows a map of North America

_Technical Requirements_
- Ruby for Good is already partnered with MapBox (https://www.mapbox.com/)
- map solution needs to be no-cost, Open Street Map could be a fallback.
- future stories will leverage data in listing to find lower zoom level to display on map

_Other Ruby4Good projects that used MapBox_
- [free forest](https://github.com/rubyforgood/freeforest)
- [terrastories is almost entirely map based](https://github.com/rubyforgood/terrastories)

lucasfeijo:
We decided to go with [MapBox](mapbox.com) because it's more cost effective and offers comparable features to other free solutions.

kevinschlabach:
From Mae in Slack -  For whoever needs MapBox credentials:
pk.eyJ1IjoibXV0dWFsLWFpZC1hcHAiLCJhIjoiY2tmZTBvd3UwMDBhbTJ4cDlic2JmMWZoaiJ9.rWscBjdl1SMT5N0yekIJYg
Lmk if an sk is needed!

thestephenmarshall:
[contributions.pdf](https://github.com/rubyforgood/mutual-aid/files/5265352/contributions.pdf)

For pin info, try and use the card design from page 2.

kevinschlabach:
per standup this morning, we should probably shift focus to the next story and proving out that data can be plotted on the map

kevinschlabach:
per standup this afternoon, this has been merged... pending deploy to demo box to review and close

kevinschlabach:
visible in demo box... not sure what the weird tile is in the middle... discuss tomorrow at standup?

<img width="1290" alt="Screen Shot 2020-09-23 at 4 18 27 PM" src="https://user-images.githubusercontent.com/44237304/94064978-76f81100-fdb8-11ea-909c-5125f21d51e2.png">

h-m-m:
[...]

PR #720, opened 2020-09-22
Contributions Page Map

danielmmack:
Screen Shot 2020-09-23 at 11 35 13 AM

This PR adds a map button and map to the Contributions page.

**References**
[707](https://github.com/rubyforgood/mutual-aid/issues/707)


danielmmack:
Just some followup questions:

- What do you guys think of the style of the map? Is there a preference or a default style that you guys had in mind?

- Is there a better place to put the style tag on line 64 in the MapBrowser.vue file?

app/javascript/pages/browse/MapBrowser.vue
```diff
@@ -0,0 +1,69 @@
...
+  },
+}
+</script>
+
+<style>
+  #map {
+    width: 100%;
+    height: 500px;
+  }
+</style>
```
exbinary, line 69: This is a great location for these styles. I don't see the `#map` element in the markup here, i assume that brought in by the `mapbox` element?

app/javascript/pages/browse/MapBrowser.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <section class="MapBrowser">
+    <mapbox
+        access-token="pk.eyJ1IjoibXV0dWFsLWFpZC1hcHAiLCJhIjoiY2tmZTBvd3UwMDBhbTJ4cDlic2JmMWZoaiJ9.rWscBjdl1SMT5N0yekIJYg"
```
exbinary, line 4: We'll probably want to pull this in via an environment variable.

That'll require reading that on the server and making it available to the front end, which we don't currently have an example of (well, we have examples of getting page data into vue, but i suspect we'll want a slightly different scheme for globals such as this).

I'm ok with punting on that for now. Would you mind leaving a `FIXME` comment in there so we come back to it (one day!).

app/javascript/pages/browse/MapBrowser.vue
```diff
@@ -0,0 +1,69 @@
...
+        :map-options="{
+          style: 'mapbox://styles/mapbox/streets-v11',
```
exbinary, line 6: I don't have a strong preference for one style over another. For reference, i found this list:
https://docs.mapbox.com/api/maps/#mapbox-styles
Of those, i'd be good with `streets`, `light`, `satellite-streets`.
Curious if others have preferences. Also easy to change, so happy to go with your suggestion.

app/views/layouts/_head.html.erb
```diff
@@ -4,6 +4,7 @@
   <%= csp_meta_tag %>
 
   <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <script src='https://api.mapbox.com/mapbox-gl-js/v1.12.0/mapbox-gl.js'></script>
```
exbinary, line 7: There are obviously tradeoffs in both directions, but there's also the option of packaging this up with webpack:
https://docs.mapbox.com/mapbox-gl-js/api/ and https://docs.mapbox.com/mapbox-gl-js/api/#mapbox-css.

I'm good with using the cdn for now.

app/javascript/pages/browse/MapBrowser.vue
```diff
@@ -0,0 +1,69 @@
...
+        @map-zoomend="zoomend"
+        @map-click:points="clicked"
+        @geolocate-error="geolocateError"
+        @geolocate-geolocate="geolocate"
+      />
+  </section>
+</template>
+
+<script>
+import Mapbox from 'mapbox-gl-vue'
```
exbinary, line 24: Awesome, thanks for finding a vue wrapper!

exbinary (approved):
So sweet! 🔥 . Some comments inline, but looking amazing 🙌🏾.

app/javascript/pages/browse/MapBrowser.vue
maricris-sn: yes, it's referenced from the mapbox module

app/views/layouts/_head.html.erb
maricris-sn: we can circle back to this. we first tried adding it via webpack but we encountered problems.

eabousaif (approved):
LGTM

Issue #718, opened 2020-09-22
Remaining issues with docker development


exbinary:
Documenting a few outstanding issues with dockerized development:

1. Running `bin/dev/test` generates the following warning:
```
ActiveRecord::EnvironmentMismatchError: You are attempting to modify a database that was last run in `development` environment.
You are running in `test` environment. If you are sure you want to continue, first set the environment using:
        rails db:environment:set RAILS_ENV=test
```
Steps to recreate: Run `bin/dev/bootstrap`, then `bin/dev/test`.

1. Could be worth silencing warnings for now. There are still a few outstanding gems (eg `mobility`) that haven't fixed their warnings generated by newer rubies.

indiebrain:
A few notes on what I _think_ I've found so far...

> ActiveRecord::EnvironmentMismatchError:

TLDR; I suspect that the `development` and `test` environments are sharing a database. 

It seems that the development `docker-compose.yml` sets a [default `POSTGRES_DB` ENV var value of `mutualaid`](https://github.com/rubyforgood/mutual-aid/blob/main/docker/development/docker-compose.yml#L3). This passes through to the application, and takes precedence in both the [`development`](https://github.com/rubyforgood/mutual-aid/blob/main/config/database.yml#L40) AND [`test`](https://github.com/rubyforgood/mutual-aid/blob/main/config/database.yml#L51) - IE development and test both use the `mutualaid` database. I currently believe this to be the cause of the `ActiveRecord::EnvironmentMismatchError`, and will put together a change set to address this.

grenewode:
@indiebrain - I think your interpretation is correct. When I wrote these, I think I was only testing it against the [`docker-development`](https://github.com/rubyforgood/mutual-aid/blob/14bb7269c30d712377d4f9d4337090203fc03f0e/config/database.yml#L42-L44) configuration. When I originally the `docker-development` configuration, the  the other configurations didn't pull values from environment variables:
https://github.com/rubyforgood/mutual-aid/blob/e10d0b05fd122130279d849b60e60ea80308caa8/config/database.yml#L24-L977
Since I didn't (and still don't) know that much about ruby configuration files, I didn't want to change the other configurations, so I created a new one for docker instead. At this point, I think you might be able to get rid of the `docker-environment` configuration altogether. It is only referenced in three places in the project. One of them is the database.yml configuration for docker your already looking at. I think others could be eliminated/unified as well:
[...]

PR #722, opened 2020-09-23
Fix environment mismatch error in docker env tests

indiebrain:
Fixes `ActiveRecord::EnvironmentMismatchError` after a
`bin/dev/bootstrap` and `bin/dev/test`.

TLDR; The `development` and `test` environments were sharing the
`mutualaid` database instead of being isolated in their own databases.

References
----------

- #718

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -27,30 +27,30 @@
 default: &default
   adapter:  postgresql
   encoding: unicode
-  username: <%= ENV.fetch('POSTGRES_USER',     nil) %>
-  password: <%= ENV.fetch('POSTGRES_PASSWORD', nil) %>
-  database: <%= ENV.fetch('POSTGRES_DB',       nil) %>
+  username: <%= ENV['POSTGRES_USER'].presence || nil %>
+  password: <%= ENV['POSTGRES_PASSWORD'].presence || nil %>
+  database: <%= ENV['POSTGRES_DB'].presence || nil %>
   # https://til.hashrocket.com/posts/6918d4e62b-use-postgresql-socket-in-databaseyml
-  host:     <%= ENV.fetch('POSTGRES_HOST',     '/var/run/postgres') %>
+  host:     <%= ENV['POSTGRES_HOST'].presence || '/var/run/postgres' %>
   # https://guides.rubyonrails.org/configuring.html#database-pooling
-  pool:     <%= ENV.fetch("RAILS_MAX_THREADS", 5) %>
+  pool:     <%= ENV["RAILS_MAX_THREADS"].presence || 5 %>
 
 development:
   <<: *default
-  database: <%= ENV.fetch('POSTGRES_DB', 'mutualaid_development') %>
+  database: <%= ENV['POSTGRES_DB'].presence || 'mutualaid_development' %>
 
 docker-development:
   <<: *default
-  database: <%= ENV.fetch('POSTGRES_DB', 'mutualaid_development') %>
+  database: <%= ENV['POSTGRES_DB'].presence || 'mutualaid_development' %>
 
 # Warning: The database defined as "test" will be erased and
 # re-generated from your development database when you run "rake".
 # Do not set this db to the same as development or production.
 test:
   <<: *default
-  database: <%= ENV.fetch('POSTGRES_DB', 'mutualaid_test') %>
+  database: <%= ENV['POSTGRES_DB'].presence || 'mutualaid_test' %>
 
 production:
   <<: *default
-  database: <%= ENV.fetch('POSTGRES_DB', 'mutualaid_production') %>
-  username: <%= ENV.fetch('POSTGRES_USER', 'mutualaid') %>
+  database: <%= ENV['POSTGRES_DB'].presence || 'mutualaid_production' %>
+  username: <%= ENV['POSTGRES_USER'].presence || 'mutualaid' %>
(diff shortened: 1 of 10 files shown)


docker/development/app/Dockerfile
```diff
@@ -11,7 +11,6 @@ ENV RAILS_ENV='' \
     POSTGRES_HOST='' \
     POSTGRES_PASSWORD='' \
     POSTGRES_USER='postgres' \
-    POSTGRES_DB='' \
```
indiebrain, line 14: An interesting caveat, if `POSTGRES_DB` is set - in this case to the empty string, [the `database.yml` uses its value rather than the default](https://github.com/rubyforgood/mutual-aid/blob/main/config/database.yml#L40). In this case when we remove the default value of `POSTGRES_DB` from the development `docker-compose.yml`, the Rails app looks for the database named `''` rather than the expected defaults of `mutualaid_development` or `mutualaid_test` depending on the environment. Furthermore, even if the value is set here to the empty value - IE `POSTGRES_DB=`, the same behavior persists. Therefore, I decided to remove this `ENV` declaration so that the expcted defaults of the `database.yml` would be used.

docker/development/docker-compose.yml
```diff
@@ -1,6 +1,5 @@
 version: "3.4"
 x-shared-postgres-environment: &x-shared-postgres-environment
-  POSTGRES_DB: "${POSTGRES_DB:-mutualaid}"
```
indiebrain, line 3: Don't use the same database for every environment. Defer to the application `database.yml` defaults instead.

grenewode: Would it be possible to set `POSTGRES_DB=''` in the docker-compose and Dockerfile files, and then change the behavior in the `database.yml` file so that it will use the default if `POSTGRES_DB` is an empty string? Then you keep the nice "inherit the expected value from the database.yml" behavior, but you can also override it with environment variables when building the image. I'm not sure if there's a standard for this in Ruby land - when I'm wearing my rather ill fitting & ad-hoc dev-ops hat, I like to do everything with environment variables. Otherwise I always seem to end up messing around with `sed` & friends trying to change some value in a configuration file somewhere.

grenewode (commented):
This looks pretty good to me - although I'm really only a casual dev-opser, and I can already tell your bash-foo is stronger than mine :wink: I do think it would be good to keep the ability to set the database with an environment variable when the docker image is built though.

docker/development/docker-compose.yml
indiebrain: > Would it be possible to set POSTGRES_DB='' in the docker-compose and Dockerfile files, and then change the behavior in the database.yml file so that it will use the default if POSTGRES_DB is an empty string?

Aye, I currently believe that this could be accomplished using Ruby's [`Object#presence`](https://api.rubyonrails.org/classes/Object.html#method-i-presence) method in the ERB interpolation inside the `database.yml`.

Mayhaps something like the following:

```yml
development:
  ...
  database: <%= ENV['POSTGRES_DB'].presence || 'mutualaid_development' %>

test:
  ...
  database: <%= ENV['POSTGRES_DB'].presence || 'mutualaid_test' %>
```

Essentially, if a  non-nil, non-empty-string value is stored in the environment, use it - otherwise, use the default.

indiebrain: Thoughts on https://github.com/rubyforgood/mutual-aid/pull/722/commits/df847ce0c48a36b4435f1eed2f73ffae6dc66b9f?

acherukuri (approved):

config/database.yml
```diff
@@ -27,30 +27,30 @@
 default: &default
   adapter:  postgresql
   encoding: unicode
-  username: <%= ENV.fetch('POSTGRES_USER',     nil) %>
-  password: <%= ENV.fetch('POSTGRES_PASSWORD', nil) %>
-  database: <%= ENV.fetch('POSTGRES_DB',       nil) %>
+  username: <%= ENV['POSTGRES_USER'].presence || nil %>
+  password: <%= ENV['POSTGRES_PASSWORD'].presence || nil %>
+  database: <%= ENV['POSTGRES_DB'].presence || nil %>
```
exbinary, line 32: To make sure i'm tracking: we're switching to `presence` because we want to ignore blank strings, whereas `fetch` would only ignore `nil`s, correct?

exbinary (approved):
Awesome! Thanks for looking into this 🙏🏾.
TBH, i'm still not sure the exact sequence of how all the pieces fit together, but i trust you and @grenewode on this 🧙🏾‍♂️.

config/database.yml
indiebrain: 💯That is correct. 

The idea  is that `presence` will return the contents of the ENV var when the ENV var contains what Ruby considers a truth-y value:

```ruby
# WUMPUS=1 
ENV['WUMPUS'].presence 
# => "1"

# WUMPUS=found_me 
ENV['WUMPUS'].presence  
# => "found_me"

# WUMPUS=true 
ENV['WUMPUS'].presence  
# => "true"

# WUMPUS=false
ENV['WUMPUS'].presence  
# => "false"
```

but will return `nil` when the ENV var is unset, or contains a false-y value:

```ruby
# WUMPUS=
ENV['WUMPUS'].presence  
# => nil

# WUMPUS=''
ENV['WUMPUS'].presence  
# => nil
```

indiebrain:
Thanks all, for the great conversations, and guidance.

PR #723, opened 2020-09-23
Silence Ruby 2.7.0 deprecation warnings

indiebrain:
This is intended temporarily silence deprecation warnings emitted from
third-party gems that have yet to become compatible with Ruby 2.7.0.
This is somewhat dangerous if left in long-term as warnings use of
soon-to-be-removed language features by the application itself will be
silenced.

This should be removed before the next time this code base upgrades
Ruby.

**NOTE**

This does NOT silence Rails deprecation warnings, these
should probably be addressed independently - a quick eyeball survey
shows that most of the Rails deprecation warnings are addressable in
the application code.

References
----------

- https://github.com/rubyforgood/mutual-aid/issues/718

### DIFF
--- a/.env
+++ b/.env
@@ -25,3 +25,13 @@ HOURS_PASSWORD_RESET_LINK_EXPIRES_AFTER=8
 # with `public_repo` access and set it in .env.local (NOT HERE)
 # https://github.com/release-it/release-it/blob/master/docs/github-releases.md
 GITHUB_TOKEN=not-a-real-token
+
+# TODO: DELETE ME - This is intended temporarily silence deprecation
+# warnings emitted from third-party gems that have yet to become
+# compatible with Ruby 2.7.0.  This is somewhat dangerous if left in
+# long-term as warnings use of soon-to-be-removed language features by
+# the application itself will be silenced.
+#
+# This should be removed before the next time this code base upgrades
+# Ruby.
+RUBYOPT=-W:no-deprecated
--- a/docker/development/docker-compose.yml
+++ b/docker/development/docker-compose.yml
@@ -6,6 +6,7 @@ x-shared-postgres-environment: &x-shared-postgres-environment
   POSTGRES_PASSWORD: "${POSTGRES_PASSWORD}"
 
 x-shared-environment-defaults:
+  - &RUBYOPT "${RUBYOPT}"
   - &APP_PORT "${APP_PORT:-3000}"
   - &EMAIL_PORT "${EMAIL_PORT:-1080}"
   - &WEBPACKER_DEV_SERVER_PORT "${WEBPACKER_DEV_SERVER_PORT:-3035}"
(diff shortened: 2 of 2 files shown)


evertoncunha (approved):

acherukuri (approved):

.env
```diff
@@ -25,3 +25,13 @@ HOURS_PASSWORD_RESET_LINK_EXPIRES_AFTER=8
...
+
+# TODO: DELETE ME - This is intended temporarily silence deprecation
+# warnings emitted from third-party gems that have yet to become
+# compatible with Ruby 2.7.0.  This is somewhat dangerous if left in
+# long-term as warnings use of soon-to-be-removed language features by
+# the application itself will be silenced.
+#
+# This should be removed before the next time this code base upgrades
+# Ruby.
+RUBYOPT=-W:no-deprecated
```
exbinary, line 37: Ah, as i think more about this, maybe a better balance is letting dev's turn this off locally via `.env.local`. The comment above is very helpful so could maybe stay here, with the the actual setting commented out, and with instructions pointing to the local override. The code below is still necessary and great to allow for the local opt-out.
Thoughts?

indiebrain: I don't think docker-compsose will use the `.env.local` by default when running the image - `.env.local` is primarily used by the `dotenv` gem when the Rails application loads, but `RUBYOPTS` needs to be set in the environment before Ruby loads - or at least that's my current understanding. 

By default [docker-compose only auto-magically loads the `.env` file into a container's environment](https://docs.docker.com/compose/environment-variables/#the-env-file). So the order of events at container creation, as I understand it, are:

1. `docker-compose` creates the container, and loads the content of `.env`
1. `docker-compose` initializes the Ruby process. During initialization, the Ruby process inspects the content of the ENV `RUBYOPTS`
1. The application is loaded, and bundler loads the application gems - including `dotenv`
1. `dotenv` loads the content of `.env.local`. At this point, the `RUBYOPTS` in `.env.local` is too late to influence the runtime of the Ruby process.
1. ... more ruby stuff.

For example, this is the behavior on my local when `RUBYOPT` is absent from `.env` and present in `.env.local`:

```sh
[maximilian: mutual-aid] silence-ruby-deprecation-warnings * →
$ cat .env
...
# RUBYOPT=-W:no-deprecated

[maximilian: mutual-aid] silence-ruby-deprecation-warnings * →
$ cat .env.local
...
RUBYOPT=-W:no-deprecated

[maximilian: mutual-aid] silence-ruby-deprecation-warnings * →
$ bin/dev/bootstrap
...
WARNING: The RUBYOPT variable is not set. Defaulting to a blank string.

[maximilian: mutual-aid] silence-ruby-deprecation-warnings * →
$ bin/dev/test
...
WARNING: The RUBYOPT variable is not set. Defaulting to a blank string.
...tons of ruby deprecation warnings...
```

I'd have to think a bit more on HOW to let `RUBYOPTS` be locally overridden.

exbinary: Oh i see! Thank you, this is very helpful learning. I had no idea `.env` was something docker recognized so was confusing myself about the sequence of operations. I really appreciate the detailed explanation.
In light of all that, i vote we merge as is, and i'll create an issue in our backlog to revisit this in 6 months or so.

PR #726, opened 2020-09-23
Remove unused docker-development environment config

indiebrain:
The local development setup with docker uses the `development` and
`test` environments and the `docker-development` environment is no
longer used. This change set removes the vestigial
`docker-development` environment configuration.

References
----------

- https://github.com/rubyforgood/mutual-aid/issues/718


ansible/files/database.yml
```diff
@@ -1,50 +0,0 @@
-# PostgreSQL. Versions 9.3 and up are supported.
```
indiebrain, line 1: Not entirely sure what context Ansible runs under, but this "replace it with a symlink" assumes that Ansible runs from the project root. This assumption feels potentially dangerous, but figured I'd include it here as a response to [this conversation](https://github.com/rubyforgood/mutual-aid/issues/718#issuecomment-697059025). Would also be fine reverting this particular bit and duplicating the config between this and `config/database.yml`.

exbinary: I'm not really qualified to know whether this will work, and don't have an ansible rig handy to test it out. I'm good with the symlink since it will (hopefully) break in obvious ways and can be easily replaced with a duplicate of `config/database.yml`.

exbinary (approved):
LGTM! That said, i don't know bupkis about ansible 😀 .

PR #730, opened 2020-09-23
P2P matching & Form

eabousaif:
Enables peer to peer communication hidden between a feature toggle. The flow allows someone to claim either an offer or an ask and get in touch with the other party.

Screen Shot 2020-09-24 at 3 39 25 PM

Screen Shot 2020-09-24 at 3 39 54 PM

Screen Shot 2020-09-24 at 3 43 07 PM


.gitignore
```diff
@@ -60,3 +60,5 @@ yarn-debug.log*
 
 # Ignore vim swapfiles
 # .*.swp
+
+vendor/bundle
```
indiebrain, line 64: @acherukuri , this is _might_ be a vestige of your PowerHRG bundle cache config - it might not be relevant to this project. Had you considered checking `~/.bundle/config` to see if `BUNDLE_PATH` is set globally to `vendor/bundle`? Most projects that aren't PowerHRG projects will let bundler install gems into the GEM_PATH instead of providing a custom `BUNDLE_PATH` as PowerHRG does.

app/models/browse_filter.rb
```diff
@@ -39,7 +39,10 @@ def contributions
   def options
     return {} unless context
 
-    { respond_path: ->(id) { context.respond_contribution_path(id)} }
+    {
+      respond_path: ->(id) { context.respond_contribution_path(id)},
+      view_path: ->(id) { context.contribution_path(id)}
```
indiebrain, line 44: `view_path` probably only need to be rendered if the `peer_to_peer_matching` feature is turned on. IE we don't want [this conditional](https://github.com/rubyforgood/mutual-aid/blob/1882769d7dd6b7137531c962978e1857ba2465eb/app/javascript/pages/browse/TileListItem.vue#L40) to be true unless P2P is turned on

app/controllers/contributions_controller.rb
```diff
@@ -15,6 +15,26 @@ def index
...
+    )
+  end
+
+  def claim_ask_form
+    contribution_id = params[:id]
+    render locals: { contribution_id: contribution_id }
+  end
+
+  def claim_ask
+    redirect_to contribution_path(params[:id])
```
indiebrain, line 35: This non-RESTful route handles a `POST` verb, but it doesn't look like it creates any resources. Was this intended to create a `Match`?

eabousaif: It is. We just wanted to get what we have so far up for review.

app/views/contributions/show.html.erb
```diff
@@ -0,0 +1,11 @@
+<section class="section">
+  <h2>Request for Contribution</h2>
+  <h1 class="title"><%= contribution.type %>: <%= contribution.all_tags_unique.map(&:humanize).join(", ") %></h1>
+  <br>
+  <div class="box">
+    <p><strong><%= contribution.person.anonymize_name %></strong> - <%= contribution.person.anonymize_email %></p> 
+    <p class="subtitle"><%= contribution.description %></p>
+    <%= button_to "Claim this Ask", claim_ask_form_contribution_path(contribution), class: "button is-primary", method: :get %>
```
indiebrain, line 8: As far as I can tell, an `Ask` is but one type of "contribution". I believe this bit is supposed to render differently based on the type of contribution. Additionally, the role of the person clicking interacting with the contribution changes depending on the type of contribtion. For example, if the contribution is an `Offer`, the user will be a `receiver` on the match, however if the contribution is an `Ask` the user will be a `provider`.

app/controllers/contributions_controller.rb
```diff
@@ -15,6 +15,26 @@ def index
...
+      locals: {
+        contribution: contribution,
+      }
+    )
+  end
+
+  def claim_ask_form
+    contribution_id = params[:id]
+    render locals: { contribution_id: contribution_id }
+  end
```
indiebrain, line 32: Should requests to this have an accompanying example?

app/controllers/contributions_controller.rb
```diff
@@ -15,6 +15,26 @@ def index
...
+
+  def claim_ask
```
indiebrain, line 34: Should requests to this have an accompanying example?

app/views/contributions/show.html.erb
eabousaif: Good catch! I can make that fix

.gitignore
acherukuri: @indiebrain, I have installed gems to `vendor/bundle` on purpose as I'm using VS Code and it is very good with indexing ruby gems from `vendor/bundle` folder. It's handy for browsing code and documentation.

terryfinn: @acherukuri You should have that in your personal global git ignore.

acherukuri: @terryfinn, Thanks for reminding that. I have totally forgotten about it.

db/migrate/20200922151556_add_peer_to_peer_matching.rb
```diff
@@ -0,0 +1,5 @@
+class AddPeerToPeerMatching < ActiveRecord::Migration[6.0]
+  def change
+    add_column :system_settings, :allow_peer_to_peer_matching, :boolean, null: false, default: false
+  end
+end
```
exbinary, line 5: Oops, we should've given you this information up front, but there's an existing setting for this, stored in the [`exchange_type`](https://github.com/rubyforgood/mutual-aid/pull/730/files#diff-1acd2e7e27a227829d5d14a91c863bb6R429) column. The [existing options](https://github.com/rubyforgood/mutual-aid/blob/main/app/models/system_setting.rb#L3-L9) support a range of exchange types, but i think for now it should be sufficient to check `System_setting.peer_to_peer?`.

Apologies for not catching this earlier! 😬

app/views/system_settings/_form.html.erb
```diff
@@ -4,6 +4,7 @@
   <div class="form-inputs">
     <%= f.input :exchange_type, as: :select, collection: @exchange_types %>
     <%= f.input :separate_asks_offers, as: :radio_buttons, label: "Separate ask and offer forms" %>
+    <%= f.input :allow_peer_to_peer_matching, as: :radio_buttons %>
```
exbinary, line 7: See comment in the corresponding migration.

app/models/system_setting.rb
```diff
@@ -32,6 +32,9 @@ def self.current_settings
     self.last
   end
 
+  def self.allow_peer_to_peer_matching?
+    current_settings&.allow_peer_to_peer_matching
+  end
```
exbinary, line 37: See comment in corresponding migration.

exbinary (commented):
First of a multi-part review.

config/routes.rb
```diff
@@ -27,12 +27,14 @@
...
   get "/thank_you", to: "contributions#thank_you", as: "contribution_thank_you"
-  resources :contributions, only: [:index] do
+  resources :contributions, only: [:index, :show] do
     member do
       get "/respond", to: "contributions#respond", as: "respond"
       get "/triage", to: "contributions#triage", as: "triage"
       patch "/triage", to: "contributions#triage_update"
       post "/triage", to: "contributions#triage_update"
+      post :claim_contribution
+      get :claim_contribution_form
```
exbinary, line 37: I think this could be better modeled as a nested resource pointing at a new controller:
```
resources :contributions, only: [:index, :show] do
  member do
    # snipped ...
  end
  
  resources :claims, only: [:new, :create]
```
`claim_contribution_form` would then map to the `new` action and `claim_contribution` to `create`.

I realize we had already gone down the non-resourceful route (pun intended! :) with `/triage` and `/respond`, so quite understandable that y'all followed that pattern. OTOH, i do think the resourceful version is a better separation of responsibilities and helps keep the `ContributionsController` small and coherent. It also sets us up better if in the future we need to, eg, revoke a claim (which would map nicely to the destroy action).

If we don't have time to refactor in this direction, that's definitely fine!

exbinary (commented):
Second of a multi-part review (routes).

db/migrate/20200922151556_add_peer_to_peer_matching.rb
acherukuri: Not a problem. I have made necessary changes and using `SystemSetting.current_settings&.peer_to_peer?` now.

config/routes.rb
acherukuri: I like this idea. I have moved all the logic to `ClaimsController` now.

app/models/browse_filter.rb
acherukuri: Thanks for reminding that. I took care of that.

app/controllers/contributions_controller.rb
acherukuri: Moved this method to `claims_controller` which is a nested resource of `contributions` And creating a match there.

app/controllers/contributions_controller.rb
acherukuri: Moved this method to `claims_controller` which is a nested resource of contributions and it is RESTful now. do we still need an example?

app/models/communication_log.rb
```diff
@@ -9,10 +9,10 @@ class CommunicationLog < ApplicationRecord
 
   scope :needs_follow_up, ->(boolean=nil){ where(needs_follow_up: boolean || true) }
 
-  def self.log_submission_email(email_object, delivery_status, submission, delivery_method=nil, current_user=nil)
+  def self.log_email(email_object, delivery_status, person, delivery_method=nil, current_user=nil)
```
exbinary, line 12: Nice change and reuse!

app/mailers/peer_to_peer_match_mailer.rb
```diff
@@ -0,0 +1,13 @@
+class PeerToPeerMatchMailer < ApplicationMailer
+  def peer_to_peer_email(contribution, peer_to_peer_match_params)
```
exbinary, line 2: Minor: the second arg could be expanded into keyword args:
```suggestion
  def peer_to_peer_email(contribution, peer_alias:, message:)
```

app/views/peer_to_peer_match_mailer/peer_to_peer_email.html.erb
```diff
@@ -0,0 +1 @@
+<%= content_tag(:div, @email_body) %>
```
exbinary, line 1: IIUC, `@email_body` is composed entirely of the message entered by the claimier on the [claim form](https://github.com/rubyforgood/mutual-aid/pull/730/files#diff-9aaded2bcd601dbec86b952e49d33003R14), right?

I suspect we'll need to add some more text around their message, explaining why the claimee is getting this email, what their next step is, how to get help etc.

Do you have thoughts on this @maebeale .

Also fine to merge as is, since this is easily updated.

exbinary (commented):
Third of multi-part review (mailer)

app/models/communication_log.rb
maebeale: we were just talking about this!

app/views/peer_to_peer_match_mailer/peer_to_peer_email.html.erb
maebeale: yup! that's my recollection also -- we have the body match the email body

agree on that text, and, the merging as is and refining afterward.

app/models/person.rb
```diff
@@ -54,6 +68,33 @@ def offer_tag_list
...
+    return if name.blank?
+
+    names = name.split(" ")
+    names.map do |n|
+      initial = n.first
+      n = '*' * n.length
+      initial + n[1..-1]
+    end.join(" ")
+  end
+
```
exbinary, line 97: This logic isn't part of the Person object's core responsibility so, adhering to the SPR, i'd love to get them extracted into a utility module.

Maybe something like `app/lib/anonymize.rb`, so that they can be used like this:

```ruby
def anonymized_name_and_email
  "#{Anonymize.name(name)} #{Anonymize.email(email)"
end
```

Note: in a departure from rails norms, i personally don't like using mixins for such cases. Happy to say more about that if anyone's interested :).

This would also make these methods very easy to test.

app/models/match.rb
```diff
@@ -41,6 +41,23 @@ def self.follow_up_status(follow_up_status)
     result
   end
 
+  def self.create_match_for_contribution!(contribution, current_user)
```
exbinary, line 44: Could we add the following comment above this method:
`# FIXME: extract into an interactor`.

I think this kind of use-case specific logic fits nicely into the interactor pattern which, among other benefits, helps to keep models from getting bloated. Unfortunately we don't have an example in the codebase i can point at currently (though we have pulling in the `active_interaction` gem), so probably makes sense to refactor this later.

app/models/match.rb
```diff
@@ -41,6 +41,23 @@ def self.follow_up_status(follow_up_status)
...
+    Match.create!(match_params)
+  end
+
+  def self.create_offer_for_ask!(ask, current_user)
+    Offer.create!(person: current_user.person, service_area: ask.service_area)
+  end
+
+  def self.create_ask_for_offer!(offer, current_user)
+    Ask.create!(person: current_user.person, service_area: offer.service_area)
+  end
```
exbinary, line 59: @maebeale , is this sufficient info on the auto-created Ask|Offer? Should we add the category as well?

app/models/person.rb
```diff
@@ -22,6 +22,20 @@ class Person < ApplicationRecord
 
   validate :preferred_contact_method_present!
 
+  def self.create_from_peer_to_peer_params!(current_user, peer_to_peer_params)
```
exbinary, line 25: Minor preference for expanding `peer_to_peer_params` into keyword arguments.

Also, please add the same fixme comment as above.

app/mailers/peer_to_peer_match_mailer.rb
acherukuri: @exbinary, Thanks for the tip. There was no point in passing all the params here. Made necessary changes here.

exbinary (commented):
Fourth of multi-part review (models).

app/controllers/claims_controller.rb
```diff
@@ -0,0 +1,44 @@
+class ClaimsController < ApplicationController
+  def new
+    contribution = Listing.find(params[:contribution_id])
+    if contribution.person&.email.blank?
+      flash.now[:alert] = "We are sorry, this contributor hasn't provided
+                         an email address yet and can't communicate this way".squish
+    end
+    render locals: { contribution: contribution }
```
exbinary, line 8: Interesting use of an explicit `render` with `locals` instead of relying on setting instance variables.

Its a little more typing but i do like the explicit definition of what gets passed to the view, and have always disliked the magic ivars in templates!

Is this a norm y'all at PHR follow? I'm definitely ok with it but would like to hear your thoughts @h-m-m, @maebeale and @svileshina. We don't use this form in any of our existing controllers. What do you think? (Just to be clear what i'm talking about, in our code base we would drop line 8 above and change line 3 to be `@contribution = ...`.)

app/controllers/claims_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def create
+    # TODO: Need to handle race conditions to prevent creating multiple matches for same contribution.
```
exbinary, line 12: Great catch! I've documented this in an [issue](https://github.com/rubyforgood/mutual-aid/issues/736).

app/controllers/claims_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    contribution = Listing.find(params[:contribution_id])
```
exbinary, line 14: Another fixme comment please to extract an interactor 🙏🏾.
In this case to keep controllers focussed.

app/controllers/claims_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+      Match.create_match_for_contribution!(contribution, current_user)
+      contribution.matched!
+    end
+    notify_peer_and_log_communication!(contribution)
+    redirect_to contribution_path(params[:contribution_id]), notice: 'Claim was successful!'
+  rescue
+    flash.now[:alert] = 'There was an error. Please try again.'
+    render :new, locals: { contribution: contribution }
+  end
```
exbinary, line 29: I think i'd lean towards letting this fail fast and relying instead on a global error handler (existing [issue](https://github.com/rubyforgood/mutual-aid/issues/734)).

app/controllers/claims_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    ActiveRecord::Base.transaction do
+      if current_user.person.blank?
+        Person.create_from_peer_to_peer_params!(current_user, name: peer_to_peer_match_params[:peer_alias],
+                                                preferred_contact_method_id: peer_to_peer_match_params[:preferred_contact_method_id],
+                                                contact_info: peer_to_peer_match_params[:preferred_contact_details])
+      end
+      Match.create_match_for_contribution!(contribution, current_user)
+      contribution.matched!
+    end
+    notify_peer_and_log_communication!(contribution)
```
exbinary, line 24: 💯

app/views/contributions/show.html.erb
```diff
@@ -0,0 +1,15 @@
+<section class="section">
+  <h2>Request for Contribution</h2>
+  <h1 class="title"><%= contribution.type %>: <%= contribution.all_tags_unique.map(&:humanize).join(", ") %></h1>
```
exbinary, line 3: This is nitpicky, but i'm not sure its good practice to put an `h2` before an `h1`?

app/views/claims/new.html.erb
```diff
@@ -0,0 +1,21 @@
+<section>
+  <%= content_tag(:h1, "Claim This #{contribution.type}", class: "title") %>
+  <div class="box">
+    <p class="mb-2">
+      Information you enter here will be delivered directly to the other party.
+      Remember to protect yourself and never provide sensitive personal info to
+      strangers.
+    </p>
+        
+    <%= simple_form_for :peer_to_peer_match, url: contribution_claims_path(contribution) do |f| %>
```
exbinary, line 10: I'm curious about the choice to prefix the fields here with `peer_to_peer_match` instead of `claim`? I'm fine with, was mostly curious (typically, the form is named the same as the controller).

exbinary (commented):
Fifth of multi-part review (controllers, views).

app/helpers/claims_helper.rb
```diff
@@ -0,0 +1,16 @@
...
+      current_user.person&.email
+    else
+      current_user.person&.phone
+    end
+  end
+
+  def peer_to_peer_preferred_contact_method
+    @peer_to_peer_preferred_contact_method ||= current_user.person&.preferred_contact_method
+  end
+end
```
exbinary, line 16: Some of this logic already exists in `Person.preferred_contact_info`. Moreover, i think it could be more straightforward, easier to follow, if we explicitly set locals in the controller action instead of relying on a helper to calculate them from the view.

I spiked out what this could look like in https://github.com/rubyforgood/mutual-aid/commit/b458ebe09886dfebf34920954a12d8d27891ba13. LMK what you think and feel free to cherry-pick or merge that commit if it works.

app/controllers/claims_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  def create
+    # TODO: Need to handle race conditions to prevent creating multiple matches for same contribution.
+
+    contribution = Listing.find(params[:contribution_id])
+    ActiveRecord::Base.transaction do
+      if current_user.person.blank?
+        Person.create_from_peer_to_peer_params!(current_user, name: peer_to_peer_match_params[:peer_alias],
+                                                preferred_contact_method_id: peer_to_peer_match_params[:preferred_contact_method_id],
+                                                contact_info: peer_to_peer_match_params[:preferred_contact_details])
+      end
```
exbinary, line 20: I think we might also need to handle when the current_user _changes_ their contact preference and info.
IIUC, such a change would currently be ignored.

exbinary (approved):
Last of multi-part review (helpers).

I hope all my suggestions weren't discouraging 😬 ! Please feel free to take the ones you agree with and leave the rest -- the code is already good and already provides much needed functionality!

🔥 🙏🏾.

(P.S, thanks for already turning around some of the earlier suggestions!)

app/models/person.rb
acherukuri: I would definitely like to know more about not using mixins for these scenarios :). Also, I have moved this logic to `Anonymize` class.

app/models/match.rb
acherukuri: That is really good to know. Do you mind if I try to implement `active_interaction`? I wasn't sure if you wanted to set up any conventions for `interactors`.

app/controllers/claims_controller.rb
acherukuri: Thanks for @indiebrain, he recently introduced this concept to me. I think it's a good way to inject dependencies into views and also comes in handy during refactoring too. I'm definitely open to reverting it to use ivars.

app/views/contributions/show.html.erb
acherukuri: You're absolutely right. I took care of that.

app/views/claims/new.html.erb
acherukuri: Changed this to claim and refactored the controller layer too :)

app/helpers/claims_helper.rb
acherukuri: @exbinary Thanks for that commit, I have cherry-picked it and made a minor change which uses safe nav for picking preferred_contact_method_id https://github.com/rubyforgood/mutual-aid/commit/b458ebe09886dfebf34920954a12d8d27891ba13#diff-30e7c2adf26639d802013e232d411680R10

app/controllers/claims_controller.rb
acherukuri: Yes, you're right. That change is completely ignored as of now. We're only considering email. My take would be to create a notification interface and depending on the contact preference method, we can either notify via an email or SMS.

Do you want me to make any specific changes here or to the form to restrict only to email?

acherukuri:
> Last of multi-part review (helpers).
> 
> I hope all my suggestions weren't discouraging 😬 ! Please feel free to take the ones you agree with and leave the rest -- the code is already good and already provides much needed functionality!
> 
> 🔥 🙏🏾.
> 
> (P.S, thanks for already turning around some of the earlier suggestions!)

@exbinary Nope, not at all discouraging. I'm happy I got to work on this project. Happy to help in any way I could. Moreover, I've picked up some new concepts and made a few friends :).

app/controllers/claims_controller.rb
exbinary: I like it as well and very open to adopting it. Lets keep it as is in this PR and that'll give our team a chance to discuss the convention. 🤠

app/helpers/claims_helper.rb
exbinary: 🙌🏾

app/controllers/claims_controller.rb
exbinary: We can discuss more at our 9am if you like, but i'm leaning toward only allowing email on the form for this iteration.
In the controller, if we have to fabricate a person instance, we can force email to be the preferred contact method.

In addition, it would be good to save the email address if the person already existed but in this case, maybe leave their preference alone, just updating their email field?

app/models/communication_log.rb
acherukuri: I noticed that method and figured why not.

app/models/match.rb
exbinary: Oh yes, please go ahead if you have the time and are motivated! 💯

I don't have any strong conventions in mind, the [gem](https://github.com/AaronLasseigne/active_interaction) is pretty good.

In a separate WIP branch, i've placed them in [`app/interactions`](https://github.com/rubyforgood/mutual-aid/tree/audit-email-code/app/interactions). The example there uses the `run!` form which assumes the interactor isn't trying to return any meaningful data back to the controller; there's the regular `run` form for when a result object is useful).

exbinary:
> Moreover, I've picked up some new concepts and made a few friends :).
As have i ❤️ :)
Amazing turn around time on this PR @acherukuri 🔥🚄.

app/controllers/claims_controller.rb
acherukuri: @exbinary I have made changes to restrict the preferred contact method to just email for now and also saving email to the person record if absent in the system. Let me know how that looks. 

I'm gonna branch off of this branch and start working on implementing `active_interaction`. I thought that shouldn't stop us from merging these changes.

acherukuri:
@exbinary @maebeale Merge PR button is enabled for me but I'm not sure if I can merge the PR by myself or you wanna do it? I thought I would bring up this question just in case.

app/controllers/claims_controller.rb
exbinary: Looks great 🚀 . Agree on making the interactor refactoring a separate branch 👍🏾.

exbinary:
Oh, one last thing -- don't we need to share the claimer's email with the claimee? I think we need to add that to the email body.

maebeale:
@acherukuri  agree w @exbinary's point about including the email address in the email that's sent, both other than that, :greenlightemoji: for merging it! :)

app/lib/anonymize.rb
```diff
@@ -0,0 +1,24 @@
+class Anonymize
```
maebeale, line 1: cool!

app/models/communication_log.rb
```diff
@@ -9,10 +9,10 @@ class CommunicationLog < ApplicationRecord
 
   scope :needs_follow_up, ->(boolean=nil){ where(needs_follow_up: boolean || true) }
 
-  def self.log_submission_email(email_object, delivery_status, submission, delivery_method=nil, current_user=nil)
```
maebeale, line 12: this is a great change

app/models/match.rb
maebeale: ooo. yes, we *should* add Category to the autogenerated Ask/Offer

app/views/claims/new.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <div class="box">
+    <p class="mb-2">
+      Information you enter here will be delivered directly to the other party.
+      Remember to protect yourself and never provide sensitive personal info to
+      strangers.
+    </p>
+        
+    <%= simple_form_for :claim, url: contribution_claims_path(contribution) do |f| %>
+      <%= f.input :peer_alias, as: :string, label: false, placeholder: "Alias or handle (do not use your real name here)", required: true %>
+      <%= f.input :preferred_contact_method_id, collection: Array(ContactMethod.email), label: "Preferred contact method", required: true %>
```
maebeale, line 12: 👍

maebeale (approved):
SOOOO excited about this feature. THANK YOU, folks!!!

Issue #708, opened 2020-09-21
Plot Contributions onto Map


kevinschlabach:
**Story**
As a community member, I want to see contributions with locations plotted on the map view on the contributions page, so that I can consume the listing via a geospatial layout

**How does it work now**
#707 put a map on the contributions page

**How should it work**
A map view presenting listed contributions _(removed from scope: and sets zoom level to include everything on the map)_

==========
**Notes**

_Workflow_
- logged in user navigates to http://mutual-aid-demo.herokuapp.com/contributions
- toggles to "Map View"
- right hand area shows a map community area based on data contained
-- if no data can be plotted and contained, then fall back to a generic zoom level of north america

_Technical Requirements_
- N/A

thestephenmarshall:
Mock is located [here](https://github.com/rubyforgood/mutual-aid/issues/707#issuecomment-697109318). For the map pins, use the cards from tile view.

kevinschlabach:
This was noted on the PR for this item, but to include it here for business view eyeballs...

The delivery of this story was not able to include setting the zoom level to the locale/metro of the data.  That will need to be broken out as a separate story as it needs more data pulled into the page and we ran out of time this week.

kevinschlabach:
this has been merged, but shouldn't be closed until it is visible in the demo environment and the prior comment is reviewed and potentially broken out as a new story...?

PR #742, opened 2020-09-25
Feature/add contributions to map

phylum:
Screen Shot 2020-09-25 at 10 46 33 AM

## Why
To allow contributions with locations to be plotted on the map view dynamically on the contributions page so that users can see the Asks/Offers closest to them.

[References 708 ](https://github.com/rubyforgood/mutual-aid/issues/708)

### Pre-Merge Checklist

**All these boxes should be checked off before any pull request is merged!**

- [ ] Any next steps that seem like a good ideas have been created as issues for future discussion & implementation

## How
Here's a list of APIs used for the Maps:
[mapbox-gl-vue](https://www.npmjs.com/package/mapbox-gl-vue)
[mapbox-sdk-js](https://github.com/mapbox/mapbox-sdk-js)

## Next Steps
- Build a more complete location data for Asks/Offers to include City, State, Zipcode, if privacy concerns allow
- Implement clustering so that as we zoom in/out we can see how many Asks/Offers are available
- Dynamically centering on the top-level Service Area

### Security
- FIXME for Mapbox Access Token should be addressed to pull it from an ENV variable
- Location data mentioned above in Next Steps


maebeale (approved):
LGTM, folks! WOO HOOOO!!!!! SUCH a huge addition to this app. Thank you!!!

app/javascript/pages/Browse.vue
```diff
@@ -18,7 +18,7 @@
       <section>
         <BrowserSelector :browser="browser" @clicked="browser = $event" />
       </section>
-      <component :is="browser" :contributions="activeContributions" class="row" />
+      <component :is="browser" :contributions="activeContributions" class="row" accessToken="pk.eyJ1IjoibXV0dWFsLWFpZC1hcHAiLCJhIjoiY2tmZTBvd3UwMDBhbTJ4cDlic2JmMWZoaiJ9.rWscBjdl1SMT5N0yekIJYg"/>
```
indiebrain, line 21: A series of questions:

- Is `accesToken` a credential to the maps API? 
- Is this a sensitive credential? Does this get served down to the client? Is there any amount of armoring necessary to protect it?

indiebrain: This also seems to be copy-pasta'd in `pp/javascript/pages/browse/MapBrowser.vue`. Channeling Dave Thomas and Andy Hunt: "There should exist a single authoritative source of knowledge within a system - IE Do not repeat yourself."

Had you considered the case where we need to change the Map API credential? It seems likely that we might miss one or more instances with the current design.

maebeale: @indiebrain yup, i believe we agreed to go with this path for this week, but that there's a ticket to move this out asap

maebeale:
here's the circleci error

app/javascript/pages/browse/MapBrowser.vue
```diff
@@ -2,105 +2,131 @@
...
+      },
+      zoomend(map, e) {
+        console.log('Map zoomed')
+      },
+      clicked(map, e) {
+        const title = e.features[0].properties.title
+        console.log(title)
+      },
+      geolocateError(control, positionError) {
+        console.log(positionError)
```
thestephenmarshall, line 123: I think we *might* want to remove these `console.log` statements, but I will defer to what MA desires here.

app/models/system_setting.rb
```diff
@@ -69,6 +69,7 @@ def peer_to_peer?
 #  confirmation_page_text_footer      :string
 #  confirmation_page_text_header      :string
 #  confirmation_page_text_link_header :string
+#  display_navbar                     :boolean          default(FALSE)
```
thestephenmarshall, line 72: Are these supposed to be checked in?

spec/factories/system_settings.rb
```diff
@@ -17,6 +17,7 @@
 #  confirmation_page_text_footer      :string
 #  confirmation_page_text_header      :string
 #  confirmation_page_text_link_header :string
+#  display_navbar                     :boolean          default(FALSE)
```
thestephenmarshall, line 20: Check this in or no?

thestephenmarshall (commented):
Looks pretty awesome so far. Thanks for all the hard work! There are a couple of comments to address. 🎸

spec/factories/system_settings.rb
phylum: @thestephenmarshall Looks like this was inserted automatically. We didn't add the change manually so we should probably defer to @maebeale @exbinary @h-m-m  for this.

app/javascript/pages/browse/MapBrowser.vue
```diff
@@ -2,105 +2,131 @@
...
-          show: true,
-          position: 'top-left',
-        }"
-        @map-load="loaded"
-        @map-zoomend="zoomend"
-        @map-click:points="clicked"
-        @geolocate-error="geolocateError"
-        @geolocate-geolocate="geolocate"
-      />
+      :access-token= this.accessToken
```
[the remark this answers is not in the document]
exbinary: Nitpick: i don't think `this.` is necessary here.

spec/factories/system_settings.rb
acherukuri: [Annotate](https://rubygems.org/gems/annotate/versions/2.7.1) adds these schema details for both models and factories after migrations are applied. I think it's okay to check this in unless mutual-aid folks say otherwise.

app/javascript/pages/browse/MapBrowser.vue
```diff
@@ -2,105 +2,131 @@
...
-      }
+  export default {
+    props: {
+      accessToken: String,
+      contributions: {type: Array, default: () => []},
+      filters: Object,
+    },
+    components: { Mapbox },
+    methods: {
+      loaded(map) {
```
exbinary, line 40: This looks great. On a second pass in the future, it could be worth taking a pass at extracting some of the sdk functions out into a separate module. There's a bit going on in this `loaded` method :).

app/models/system_setting.rb
exbinary: Yes, we have a gem that automatically inserts these comments and keeps them up to date with new migrations.

exbinary (approved):
Looks good to me! TBH i can't provide a meaningful review of the mapbox related code but the general shape of things look good (a couple of very minor notes inline).

app/models/system_setting.rb
danielmmack: Thanks for the clarification @exbinary !

app/javascript/pages/browse/MapBrowser.vue
phylum: `this.` is necessary because we are accessing MapBrowse's props and `{{ accessToken }}` didn't pass the token at all.

app/blueprints/service_area_blueprint.rb
```diff
@@ -1,7 +1,7 @@
 class ServiceAreaBlueprint < Blueprinter::Base
   identifier :id
 
-  fields :name, :description
+  fields :name, :description, :location
```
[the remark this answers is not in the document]
phylum: Thanks, that was a minor oversight by us while we were learning our way around Blueprint

PR #743, opened 2020-09-25
Allow ActionMailer to deliver SMTP to email docker service

### LINKED ISSUE #737: ActionMailer doesn't send email in development
In order to observe the SMTP functionality of the system
We want ActionMailer- in development - to use a local SMTP blackhole
So we can validate mail delivery without the need to actual send mail via a live SMTP instance

The application already uses `mailcatcher` as an email black hole - in fact the local `docker-compose.yml` includes a service to start `mailcatcher`. However the application does not appear to be configuring ActionMailer for local development, nor are the appropriate networking bits in place in docker to allow inbound SMTP traffic to the `mailcatcher` service. 

References
------------

- https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-configuration
- https://github.com/sj26/mailcatcher
- https://github.com/rubyforgood/mutual-aid/blob/e82ca66ae744e7366d8eaf437e559e0af6019822/config/environments/development.rb#L34-L37

indiebrain:
## Why

In order to observe the SMTP functionality of the system
We want ActionMailer- in development - to use a local SMTP blackhole
So we can validate mail delivery without the need to actual send mail via a live SMTP instance

The application already uses mailcatcher as an email black hole - in fact the local docker-compose.yml includes a service to start mailcatcher. However the application does not appear to be configuring ActionMailer for local development, nor are the appropriate networking bits in place in docker to allow inbound SMTP traffic to the mailcatcher service.

Resolves #737 

### Pre-Merge Checklist

**All these boxes should be checked off before any pull request is merged!**

- [x] Security & accesibility have been considered
- [x] Any next steps that seem like a good ideas have been created as issues for future discussion & implementation

## What

- [x] Configure ActionMailer in development
- [x] Inject SMTP configuration into the `app` service

## How

The `email` service already runs as part of `bin/dev/serve`, but the `app` service doesn't know how to connect to it. In this changeset we:

- Allow the user to remap the SMTP port of the email service (not entirely necesary, but more consistent with how the services are currently configured)
- Remap the `email` services configuration into the `app` container using the ENV var names the Rails app expects
- Configure ActionMailer to send outbound email via SMTP to the `email` service via the ENV vars injected in the step above.  

### Testing

mailcatch-with-docker

This exposes mailcatch's SMTP service on the host machine's network interface - by default at `0.0.0.0:1025`. Perhaps it would be more appropriate to only expose this service to the docker-local network?


config/environments/development.rb
```diff
@@ -36,8 +36,21 @@
...
+  config.action_mailer.smtp_settings = {
+    address: ENV['SMTP_ADDRESS'],
+    port:    (ENV['SMTP_PORT'].presence || 587).to_i,
+    domain:  ENV['SYSTEM_HOST_NAME'],
+  }
+
   # per devise's suggested settings
-  config.action_mailer.default_url_options = { host: 'localhost', port: 3000 }
+  config.action_mailer.default_url_options = {
+    host: ENV['SYSTEM_HOST_NAME'],
```
indiebrain, line 51: This is set in `.env` to localhost by default.

docker/production/app/Dockerfile
```diff
@@ -16,10 +16,6 @@ ENV RAILS_ENV='' \
     POSTGRES_USER='postgres' \
     POSTGRES_DB='' \
     SECRET_KEY_BASE='' \
-    EMAIL_PORT='' \
-    EMAIL_HOST='' \
-    EMAIL_FROM_ADDR='' \
-    EMAIL_DEFAULT_URL_HOST=''
```
indiebrain, line 22: These ENV vars aren't used by the `production` docker config, nor are they used by the application.

acherukuri (approved):

docker/production/app/Dockerfile
maebeale: @grenewode FYI ☝️

maebeale (approved):
<3 !!!

PR #749, opened 2020-09-28
Replace guard with the much lighter rerun gem

exbinary:
## Why

`guard` with [`guard-rspec`](https://github.com/alexch/rerun) is a fine solution but presented a few imperfections:
1. It would slow down over time; every automatically triggered run would take slightly longer.
1. By default it uses a 1 to 1 mapping of spec to app code. This is often a good choice, but i often want to run a _set_ of specs as i make related changes. This is difficult to achieve with guard.
1. It has a non-trivial configuration scheme. This is not often a problem since plugin gems usually generate configuration for us, but when we do need to tweak something, there's a significant api to learn.

[`rerun`](https://github.com/alexch/rerun) is a much simpler tool that addresses all of the above.

I'm not aware of any core/recent contributors using guard, but if anyone is, we can certainly keep it alongside `rerun`.
In fact the latter doesn't even need to be in the project Gemfile; there's just a slight advantage to putting such tooling in here to reduce external tooling (eg in a docker container).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,8 +51,8 @@ end
 
 group :development do
   gem 'annotate'
-  gem 'guard-rspec', require: false
   gem 'listen', '>= 3.0.5', '< 3.2'
+  gem 'rerun'
   gem 'spring'
   gem 'spring-commands-rspec', require: false
   gem 'spring-watcher-listen', '~> 2.0.0'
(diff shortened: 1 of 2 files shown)


maebeale (approved):
👍

PR #772, opened 2020-10-10
Fixes db seeds rake tasks for development

### LINKED ISSUE #719: Fix two minor bugs in `db:import_all_seeds` rake task.
Ran into a couple of errors attempting to run `rake db:import_all_seeds` into the demo env.

1. Ran into a `ContactMethod` with a `nil` `field`. I see this both in demo and local, so i suspect something in our seeding is creating that row:
```
NoMethodError: undefined method `downcase' for nil:NilClass
/app/db/seeds/dev_seeds.rb:35:in `block in <main>'
```

2. Trying to create `User` records with a too-short password:
```
ActiveRecord::RecordInvalid: Validation failed: Password is too short (minimum is 8 characters)
...
/app/db/seeds/dev_seeds.rb:171:in `block in <main>'
```

kinduff:
## Why
Closes #719, since sometimes the seed would not perform successfully.

### Pre-Merge Checklist

## What and How
In favor of having consistent environments, I took the liberty to replace the sequence `db:drop db:create db:migrate` with `db:reset db:migrate`. This ensures consistency between environments and avoids running all the migrations (this could go bad easily). More notes in c69a95a.

Besides that, I also added `db:seed` to `import_all_seeds` to always seed the base, and then the development environment, which fixes the exception regarding an undefined method, and I added an `&` in case something else happens.

In the other hand, the bug mentioned at #719 regarding password length, I added a prefix, more notes at 9aecf92.

### Testing
Didn't add any testing for this.

## Next Steps
I see a lot of opportunities to improve the seed files, both structurally and in best practices. I would also advice to upgrade to Ruby 2.7.2 to avoid some warnings we're having.

### DIFF
--- a/db/seeds/dev_seeds.rb
+++ b/db/seeds/dev_seeds.rb
@@ -168,7 +168,7 @@ end
 end
 
 Person.where.not(email: nil).sample(7).each do |person|
-  User.where(email: person.email).first_or_create!(confirmed_at: person.created_at, password: person.name.parameterize)
+  User.where(email: person.email).first_or_create!(confirmed_at: person.created_at, password: "password-#{person.name.parameterize}")
 end
 
 # autoemail logs per Listing
--- a/db/seeds/dev_seeds.rb
+++ b/db/seeds/dev_seeds.rb
@@ -32,8 +32,8 @@ state = ["NY", "MI", "DC", "NC"].sample
   Person.where(location: location,
                name: Faker::Name.name, 
                preferred_contact_method: contact_method,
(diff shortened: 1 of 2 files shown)


exbinary (approved):
Love all these changes! Thanks @kinduff 🙏🏾💥

Would also love to hear your ideas for improving seed files.

And i've added #782 to upgrade ruby.

maebeale:
@kinduff hey there! that tag isn't needed bc we have a repo-level label/tag for Hacktoberfest. (it *is* still needed if the PR isn't merged by the end of the month)

PR #774, opened 2020-10-10
Fit all the offers in the initial view for the map

### LINKED ISSUE #753: Zoom initial map view
Zoom to tightest scale that encompasses all existing contributions.

kinduff:
## Why
Fixes #753, since the initial map view was not displaying all the offers.

### Pre-Merge Checklist

Regarding the CHANGELOG, not sure if I should do it. Please advice.

## What
- Uses bounds extender with processed markers in order to have all bounds and fit the map
- Uses padding to avoid cutting off the icons (see screenshot attached)

image

## How
It uses [LngLatBounds](https://docs.mapbox.com/mapbox-gl-js/api/geography/#lnglatbounds) and it's method `extend` in order to have a bound box that later can be set to fit the map.

### Testing
N/A

## Next Steps
N/A

- [x] Not sure if I should change the CHANGELOG myself.
- [x] I'm really worried about the performance of the `contributions_controller`. There are a lot of queries to the database that we should be able to either use a JOIN, an INCLUDES, etc.

### Accessibility
N/A

### Security
N/A

### Meta
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ We're still in the [initial development phase](https://www.jering.tech/articles/
 This changelog also serves to acknowledge the incredible people who've contributed brilliance, effort and being. Their handles are listed under the first release they each  touched. 💗🙏🏾
 
 
+## [0.3.1] - 2020-10-10
+### Enhancement
+* Fit all the offers in the initial view for the map #774
+
 ## [0.3.0] - 2020-10-05
 ### Breaking changes
 * The `SMTP_ADDRESS` environment variable has been renamed to `SMTP_HOST` #750
--- a/app/javascript/pages/browse/MapBrowser.vue
+++ b/app/javascript/pages/browse/MapBrowser.vue
@@ -97,6 +97,7 @@
               },
             })
 
+            var bounds = new mapboxgl.LngLatBounds();
             geojson.features.forEach(function (marker) {
               var el = document.createElement('div');
               el.className = 'marker ' + marker.properties.contributionType.toLowerCase() + '-marker';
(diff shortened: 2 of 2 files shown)


maebeale:
👋 @kinduff ! yes please re changelog.

maebeale:
Agree re Contributions controller. We're going to be refactoring it anyway bc we need to mix in Community Resources data, so, don't think we need a new Issue for that. Is that something you'd want to take on? (Could def use the help on it!)

kinduff:
@maebeale Thanks for the clarification, the CHANGELOG is now updated. Re Contributions, would love to help! Let me know what's on your mind and will contribute for sure.

exbinary:
@kinduff , appreciate your work on this and apologies for the long delay getting to it.

The changes here look good in terms of ensuring the map _expands_ to fit existing contributions.

However they don't solve the other part of this, which is to _zoom in_ so that the initial view of the map is as narrow as possible while still fitting all contributions. This is preferable because most/all mutual aid groups serve a specific locality (a ward, city or region) and so starting with the whole U.S as a default is always going to be too wide.

So for example, if all contributions are in and around Philadelphia, then the initial view of the map should narrow in on the city. If, on the other hand all contributions are within the Shaw neighborhood of DC, then the initial view should zoom into just that neighborhood, more or less.

Does that make sense?

I'm happy to merge what you have since it does solve one part of this problem, even though not the whole issue.

kinduff:
@exbinary Hello! Just for the sake of the context and to try to understand what the missing part of the issue is.

What's missing from this PR? What I did is to calculate the existing contributions on the map and fit the view (zoom in) as better as possible, so if all contributions are just in one city, the zoom will display the city in question.

exbinary:
@kinduff, you're so right! Turns out i had fetched your branch but not switched to it when testing! 🤦 
My bad, this is perfect and i'll merge it shortly! 🙏🏾
(And i appreciate your grace in responding kindly 😇).

exbinary (approved):

PR #781, opened 2020-10-12
Remove sanitize-for-vue

exbinary:
## Why
Found a bug with single-quotes in the custom regex we'd written and while looking into this, realized we probably don't need a separate sanitizer for vue. We were able to get the only current use of this method (on the landing page) to work with sanitize_for_rails along with regular serialization (.to_json).

### Pre-Merge Checklist

## How
Refactored `HtmlSanitizer` class to drop `sanitize-for-vue` and rename `sanitize-for-rails` -> `sanitize`.

### Security
We're still running user content through `sanitize`. In fact, this is probably more secure now that we're not relying on a custom regex for sanitization.

Co-authored-by: maebeale <maebeale@gmail.com>

### DIFF
--- a/app/controllers/public_pages_controller.rb
+++ b/app/controllers/public_pages_controller.rb
@@ -27,9 +27,11 @@ class PublicPagesController < PublicController
   end
 
   def landing_page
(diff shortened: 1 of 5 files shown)


exbinary:
@h-m-m , would love your 👀  on this one.

PR #790, opened 2020-10-16
Allow for navigating to glossary without login

### LINKED ISSUE #783: Make our glossary publicly accessible
It currently lives at http://mutual-aid-demo.herokuapp.com/admin/glossary which requires login.
We'd like to make it accessible without having to login.

fridaland:
## Why
It is no longer desired to display the glossary to users that are logged in only. 

### Pre-Merge Checklist

## What
- [x] Adjust glossary route
- [x] View glossary without logging in

## How
Modify glossary route to be it's own resource rather than one
nested in the admin resource.
This means we can create a dedicated controller for Glossary as well as
add a view test to ensure we can render a new view without logging in.

### Testing
Adding the fist view test to the app! View tests come in handy in cases like these so we can confirm we are successfully rendering the appropriate views and in this case, without requiring a login. 
This particular view requires a `Listing` record, thus we create it as part of the test setup.

## Next Steps
None

Ensure routing changes are okay with maintainers

### Accessibility
No change

### Security
Glossary is now viewable by anyone

### Meta
Resolves #783 

Screen Shot 2020-10-16 at 11 11 10 AM


config/routes.rb
```diff
@@ -71,6 +70,7 @@
   resources :system_settings
   resources :teams
   resources :users
+  resources :glossary, only: [:index]
```
exbinary, line 73: An adjustment i'd suggest is to make it a [singular resource](https://edgeguides.rubyonrails.org/routing.html#singular-resources) instead. Not much would change in terms of code, but i think that feels more semantically precise.

app/controllers/glossary_controller.rb
```diff
@@ -0,0 +1,5 @@
+class GlossaryController < ApplicationController
+  before_action :authenticate_user!, except: [:index]
```
exbinary, line 2: +1 to this being its own controller 🙌🏾

I think this can extend `PublicController` instead which skips authentication.

exbinary (commented):
Awesome @fridaland !
Left a couple of minor comments.
Thanks for working on this! 🙏🏾

exbinary:
@fridaland , am tagging this as `hacktoberfest-accepted` to make sure you get credit in time.
LMK if you find time to look over the PR comments.

exbinary:
@fridaland , i'm going to merge this into a feature branch on our repo in order to resolve merge conflicts and make some small changes (as mentioned in review). This way, you'll still be credited for your work on the PR. Thanks!

Issue #537, opened 2020-06-25
Add toggle for show/hide name on Contributions page


maebeale:
Dispatchers want the ability to see Contributor's name on Contributions page

JadeDickinson:
I'm working on this one if that's ok?

exbinary:
Fixed in #791 and #917 .

PR #791, opened 2020-10-18
Show contributor name on Contributions page

JadeDickinson:
## Why
Shows contributor's name on the List and Tile views of the Contributions page (already shown on Map view).

https://github.com/rubyforgood/mutual-aid/issues/537

### Pre-Merge Checklist

- [ ] All new features have been described in the pull request

## What
- [x] Displays Contributor name on Contributions list view
- [x] Displays Contributor name on Contributions tile view

Screenshot 2020-10-18 at 10 05 38
Screenshot 2020-10-18 at 10 12 38

## How
Contributor's name is stored as (Name) along with details of the offer in the name field. I've used a regex to match for text within brackets.

I wasn't totally clear on the requirements - do we want the ability to Toggle showing contributor names as well?

If so, how would we want this to look on the page - would a toggle box belong alongside the Filters given that it's not a Filter but a display option?

### DIFF
--- a/app/javascript/pages/browse/ListBrowser.vue
+++ b/app/javascript/pages/browse/ListBrowser.vue
@@ -8,6 +8,7 @@
         <th>Service Area</th>
         <th>Connect</th>
         <th>Respond</th>
+        <th>Contributor Name</th>
 <!--        <th>Details</th>-->
       </tr>
       <tr v-for="contribution in contributions" :key="contribution.id">
@@ -30,6 +31,7 @@
         <td>
           <a :href="contribution.respond_path" class="button icon-list is-primary"><span class=""> Respond</span></a>
         </td>
+        <td>{{ contribution.name.match(/\((.*)\)/)[1] }}</td>
 <!--        <td>{{ contribution.title }}</td>-->
       </tr>
     </table>
--- a/app/javascript/pages/browse/TileListItem.vue
+++ b/app/javascript/pages/browse/TileListItem.vue
@@ -32,6 +32,9 @@
           <div v-if="service_area" class="has-text-grey-lighter">
             {{ service_area.name }}
           </div>
+          <div>
+            {{ `From: ${name.match(/\((.*)\)/)[1]}` }}
+          </div>
         </div>
       </div>
       <div class="actions">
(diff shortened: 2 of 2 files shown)


h-m-m:
Hi, @JadeDickinson, and thanks for contributing! I'm so happy you're interested in some of this project's open issues

To answer your question, I don't think we'd want a toggle on the filter page itself. We may want a system-wide toggle, or a permissions-based toggle where you only see the name if you have the permission to do so.

That permissions- or settings-based toggle, if and when we add it, would likely live in the Rails side. For that reason, maybe it's better if we split the responsibilities up. For example, we could have the Rails side either supply a name or not (or leave it blank?) and then the Vue components could display the name only when it exists. What do you think about this approach?

maebeale:
@JadeDickinson hello there! we'd love to move forward on this feature. might you have time in the near future to pick this back up, or, might you be ok with our taking it from here?

JadeDickinson:
@maebeale Hi there! I wasn't sure what the favoured approach was at the time which is why I didn't work further on it. I'm not sure I have the time to at the moment, so feel free to take it from here

maebeale:
Thank you for the quick reply, @JadeDickinson ! If/when you're able to hop back in, we'd love your contributions -- which include questions and threads like the above. Best to you!

app/javascript/pages/browse/ListBrowser.vue
```diff
@@ -30,6 +31,7 @@
         <td>
           <a :href="contribution.respond_path" class="button icon-list is-primary"><span class=""> Respond</span></a>
         </td>
+        <td>{{ contribution.name.match(/\((.*)\)/)[1] }}</td>
```
hairedfox, line 34: `/\((.*)\)/` this regex is identical with the one below. I think you should use some constant to config it somewhere, then use it for both places.

Next time when we want to change this regex, we just need to do it once.

JadeDickinson: That's a good idea - however I didn't end up merging this PR (see the discussion of the feature above). So I won't be refactoring further for now as this code isn't in main.

@svileshina has taken over work on this feature on https://github.com/rubyforgood/mutual-aid/pull/917

svileshina:
hi @JadeDickinson, we merged this alongside #917. Thank you for your work on this!

PR #792, opened 2020-10-20
Add Pagination for People

### LINKED ISSUE #528: Add pagination to People index

yez:
## Why

Closes #528

Adds pagination to the `peoples_controller#index` route. 

### Pre-Merge Checklist

## What

Using the [`pagy`](https://ddnexus.github.io/pagy/how-to#quick-start) gem,
add pagination to the `peoples_controller#index` route.

## How

Two modules have been included:

  1. `Pagy::Backend` in `ApplicationController`
  2. `Pagy::FrontEnd` in `ApplicationHelper`

The navigation for `People` has been added to the `index` view and is
present only when there are enough `People` to paginate. `pagy` takes
care of all query parameters around pagination.

people_pagination

### Testing

For testing, I introduced a new folder for controller testing and added
the `rails-controller-testing` gem to allow expectations via `assigns`.

I was not sure of any front-end specific testing for this feature but am
happy to create one if we need it.

## Next Steps

N/A

Not any at the moment.

### Accessibility

It should speed up the page and help with lower bandwith devices

### Security

N/A

### Meta

N/A


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,8 @@ class ApplicationController < ActionController::Base
   before_action :set_system_setting
   around_action :switch_locale
 
+  include Pagy::Backend
+
```
exbinary, line 9: I try not to include much in `ApplicationController` (or `ApplicationHelper`). It's so easy for these base classes to quickly become 'kitchen-sink's of clutter, all of which is inherited by all controllers.

I realize that is what's recommended in the `pagy` docs and also pretty common in the rails world, so definitely not a critique of your work!

If you're open to trying some alternatives:
* Call `Pagy::Backend.pagy` directly in actions. I _think_ that should work and is very explicit and straightforward. `Pagy::Frontend.pagy_nav` in the view. On the downside, it's more wordy :grin:
* Include the backend and frontend only in controllers that need them (+ a `helper` call might also be necessary?).
* Extract a thin module/concern to reduce the duplication of `include`ing the pagy modules in controllers 

Curious your thoughts on this?

app/views/people/index.html.erb
```diff
@@ -33,5 +33,14 @@
         <td><%= edit_button(person.user, "", "fa fa-user-edit") if person.user_id %></td>
       </tr>
     <% end %>
+    <% if @pagy.pages > 1 %>
+      <tr>
+        <td></td>
+        <td></td>
+        <td colspan=2>
+          <%= pagy_nav(@pagy).html_safe %>
```
exbinary, line 41: I noticed that pagy ships with some [bulma specific helpers](https://ddnexus.github.io/pagy/extras/bulma). Not sure exactly what that provides, but am curious.

Gemfile
```diff
@@ -47,6 +48,7 @@ end
 group :test do
   # an XML formatter is required for fancier CircleCI results
   gem 'rspec_junit_formatter'
+  gem 'rails-controller-testing'
```
exbinary, line 51: I really appreciate you taking the time to add tests! 🙌🏾

That said, i'm quite hesitant to add this gem since the rails team very intentionally [deprecated and removed](https://github.com/rails/rails/issues/18950) access to `assigns`.

Relatedly, we've been preferring [request specs](https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec) over controller specs. My impression is that is becoming the recommended best practice.

I'd love to hear your thinking on all this?

exbinary (commented):
Awesome @yez , thank's for working on this! 🚀
Looks like it was pretty straightforward to get `pagy` working, which is a good sign!
I commented about a couple of design questions, but nothing major.
LMK your thoughts.

app/controllers/application_controller.rb
yez: I am open to calling `Pagy::Backend.pagy` explicitly, that's no problem. I originally had the `include` only in the `PeopleController` but thought that more pagination was likely coming soon so I'd get ahead of it (a pre-mature optimization on my part). 

I'll see if the backend and frontend direct calls work the same and then update if they do.

Gemfile
yez: I can see the hesitation towards `assigns` and the `rails-controller-testing` gem. I introduced them because what I care about testing here wasn't so much the view rendering but if the count of `people` is what I'd expect. 

Without the gem I could do the same test with an assertion against `controller.instance_variable_get(:@people)` but I'm not sure that's much better (reaching into the instance variables of an object without a reader is smelly code). 

I could move this to a request spec and assert that the expected number of `<td>`s exist or `<td>`s with the expected people's names in them.

app/views/people/index.html.erb
yez: I've not used `bulma` before so I'm happy to try out the `bulma` specific helpers. I'll see if one is obviously best.

app/controllers/application_controller.rb
yez: Looks like calling the `pagy` method directly on the `Pagy::Bakend` module doesn't work, we have to include it. I'll move the include to the `people_controller` instead. Same thing for `Pagy::Frontend`, I'll move that to a new `people_helper` to reduce the impact on the rest of the app.

yez:
@exbinary I have addressed all your comments. I think the request spec might be a little slow with all the matching, so I'm open to suggestions there.

spec/requests/people_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      get people_url
+      expected, remainder = people.partition.with_index { |p, i| i < Pagy::VARS[:items] }
+
+      expected.each do |person|
+        expect(response.body).to match(/#{person.name}/)
+      end
+
+      remainder.each do |person|
+        expect(response.body).to_not match(/#{person.name}/)
+      end
```
exbinary, line 26: I played around with a variation that counts the number of people:
```suggestion
      people_links = response.body.scan %r{href="/people/\d+/.+"}
      expect(people_links.size).to eq Pagy::VARS[:items]
```
But it didn't really speed things up much (both are coming in ~1.9s).

I think the people creation actually takes up a chunk of that time (~.5s), which i imagine would be incurred with a controller test as well.

This is probably part of the overhead with request specs :(

app/controllers/people_controller.rb
```diff
@@ -1,6 +1,8 @@
 class PeopleController < ApplicationController
   before_action :set_person, only: [:show, :edit, :update, :destroy]
 
+  include Pagy::Backend
+
```
exbinary, line 5: Thanks for pulling this out, feels better!

spec/requests/people_spec.rb
yez: In that case, do you think leaving it like it is would be fine?

app/helpers/people_helper.rb
```diff
@@ -0,0 +1,3 @@
+module PeopleHelper
+  include Pagy::Frontend
+end
```
exbinary, line 3: What do you think about a concern that handles both the backend and frontend inclusion, eg:
```ruby
# app/controllers/concerns/pagination.rb
module Pagination
  extend ActiveSupport::Concern

  include Pagy::Frontend
  include Pagy::Backend

  included do
    helper_method :pagy_bulma_nav
  end
end
```

app/views/people/index.html.erb
```diff
@@ -33,5 +33,14 @@
         <td><%= edit_button(person.user, "", "fa fa-user-edit") if person.user_id %></td>
       </tr>
     <% end %>
+    <% if @pagy.pages > 1 %>
+      <tr>
+        <td></td>
+        <td></td>
+        <td colspan=2>
+          <%= pagy_bulma_nav(@pagy).html_safe %>
```
exbinary, line 41: Cool, looks more bulma-ish :)

app/helpers/people_helper.rb
yez: Oh that's good, I'll do this.

exbinary (approved):
This is great! Appreciate you @yez 💯 🙏🏾

PR #793, opened 2020-10-20
Add rubocop

### LINKED ISSUE #34: Add rubocop

smoline:
## Why
This adds rubocop to the project, leaving the rubocop corrections for other PRs. Rubocop needed to be added to the project before the project progressed any further so some standard linting and best coding practices could be followed for Ruby.

Closes #34 Add rubocop

### Pre-Merge Checklist

## What
[x]  rubocop-rails gem
[x] .rubocop.yml

## How
The gem was installed through the Gemfile, and the config file was added to the root of the project.

### Testing
I tested it out by using it and found over 3300 suggested corrections, I then modified the .rubocop.yml file to exclude standard files that don't need linting and brought that number down to 2400.

## Next Steps
The next step would include creating a PR that solves a bunch of the rubocop warnings and corrections.

If a ticket could be created to address some rubocop corrections and I be assigned to it, I would appreciate it. I will start work on it immediately.


.rubocop.yml
```diff
@@ -0,0 +1,37 @@
...
+    - 'db/**/*'
+    - 'bin/**/*'
+    - 'Gemfile'
+    - 'Rakefile'
+    - 'spec/spec_helper.rb'
+    - 'node_modules/**/*'
+    - 'tmp/*'
+    - 'config/initializers/**/*'
+    - 'lib/tasks/*'
+    - 'Guardfile'
```
exbinary, line 16: I'm curious what your criteria for exclusions were?

Some of these, i would love to have `rubocop` weigh in on:
* `config/initializers` because i'm not sure we should be doing anything complicated in there
* `lib/tasks` because some rake tasks can get pretty unwieldy if they aren't kept in check!
* `spec_helper` is another one that can get bloated

Others i don't feel a particular need to have coverage for, but am curious what about them was a problem?
* `db/`
* `bin/`
* `Gemfile`

Lastly, there are a few i _think_ we can drop from this list:
* `node_modules` is excluded by [default](https://github.com/rubocop-hq/rubocop/blob/master/config/default.yml)
* Same for `tmp`
* I don't think our project has a `Rakefile`
* We _do_ have a `Guardfile`, but we shouldn't -- that was an oversight on my part when i removed the guard gem 🤭 

Sorry if that is overly picky! None of it is a major concern, for sure.

.rubocop.yml
```diff
@@ -0,0 +1,37 @@
...
+
+Layout/LineLength:
+  Max: 100
```
exbinary, line 19: I personally don't mind the default line length of `120` but am very open to other opinions on that.

.rubocop.yml
```diff
@@ -0,0 +1,37 @@
...
+  Exclude: 
+    - 'config/routes.rb'
```
exbinary, line 21: I'd love to keep `routes.rb` in the fold. Looks like there's only one line currently exceeding `100`.

.rubocop.yml
```diff
@@ -0,0 +1,37 @@
...
+
+Metrics/BlockLength:
+  Exclude:
+    - config/**/*
+    - spec/**/*
```
exbinary, line 26: I wonder if using the [ExcludedMethods attribute](https://docs.rubocop.org/rubocop/1.0/cops_metrics.html#configurable-attributes-2) of this cop would give us more fidelity? If so, i'd like to keep `config` and `spec` in the fold.

.rubocop.yml
```diff
@@ -0,0 +1,37 @@
...
+
+Lint/AmbiguousBlockAssociation:
+  Exclude:
+    - spec/**/*
```
exbinary, line 30: This one i'm not as familiar with and am curious what about specs makes them not conform?

exbinary (commented):
Thanks for adding this @smoline ! 🚀🙏🏾
Had a few questions and comments; nothing major.
Will look at #795 next!

.rubocop.yml
smoline: sounds good to me

.rubocop.yml
smoline: That would clear up a bunch of those offenses, I am open to changing this.

.rubocop.yml
smoline: I just pulled some standard .rubocop.yml files and used the ones that I have seen the most consistently. I have no issues with changing anything. No worries!

.rubocop.yml
smoline: no problem

.rubocop.yml
smoline: I will have to pull that one out and see what it adds in offenses. I am happy to change it.

.rubocop.yml
smoline: We do have a Rakefile though, but like I said I am fine changing anything.

smoline: node_modules are not excluded by default, I put that one in there myself.

exbinary (approved):
Looks great!  Thanks for working on this, and the adjustments, @smoline ! 🙏🏾

PR #803, opened 2020-11-03
Extracts P2P logic to interactions

acherukuri:
## Why

As part of [P2P implementation](https://github.com/rubyforgood/mutual-aid/pull/730/files), all the business logic was baked into the controller layer. This is an effort to extract all the business logic to appropriate active interactions. More changes to come.
### Pre-Merge Checklist

- [ ] All new features have been described in the pull request


app/interactions/email_peer.rb
```diff
@@ -0,0 +1,25 @@
...
+      peer_alias: peer_alias,
+      message: message,
+    )
+
+    status = Messenger.new(peer_to_peer_email, 'peer_to_peer_email').deliver_now
+
+    CommunicationLog.log_email(
+      email: peer_to_peer_email,
+      delivery_status: status,
+      person: contribution.person, # to person?
```
acherukuri, line 21: @exbinary, Can you please confirm if the below assumption is right?

`CommunicationLog#person` - the person to whom the email was sent (whoever created the contribution)
`CommunicationLog#initiator` - the person who initiated the P2P flow

maebeale: Idk if the #person will *always* the person who created the Contribution. (Ideally later people will be able to reply to an initiator even if that person never created the "matching/opposite" Contribution. At this point I think we're automagically creating those matched items, so there still will be a "creator", tho idk if we're definitely setting "created_by" when we do that...)

We might want to change "person" to "recipient" and "initiator" to "sender" ?

maebeale: But, yeah, I wouldn't think the Person should be the current_user.person ???

acherukuri:
@exbinary Can you provide the link to the issue if one was logged already for this work? Ignore if that's not necessary.

app/interactions/email_peer.rb
acherukuri: @maebeale I think we should rename `person` to `recipient` and `initiator` to `sender` as well (those are more semantic). 

When we initially implemented the P2P, we were setting both `person` & `initiator` attributes to [`current_user.person`](https://github.com/rubyforgood/mutual-aid/pull/730/files#diff-2d74dcaa2d57cb4ea0e3178afebb30be9726cb4995e0f158fca886ae7901c4c0R45). When I was refactoring the code, that struck as odd as we need to save both recipient & sender details but we were saving just the sender's details.

maebeale: yes, absolutely, @acherukuri ! i'm not sure how we ended up w where we were, but i def agree w your instincts. we will have both people/accounts at that point, right?

acherukuri: Yes, we'll have both account details then.

app/interactions/create_match_for_contribution.rb
```diff
@@ -0,0 +1,38 @@
...
+    Match.create!(match_params.merge(status: 'match_confirmed'))
+  end
+
+  def create_offer_for_ask!
+    Offer.create!(counter_contribution_params)
+  end
+
+  def create_ask_for_offer!
+    Ask.create!(counter_contribution_params)
+  end
```
exbinary, line 29: I think this can be simplified somewhat:
```suggestion
  def execute
    Match.create! match_receiver_and_provider.merge(status: 'match_confirmed')
    contribution.matched!
  end

  private

  def match_receiver_and_provider
    if contribution.ask?
      { receiver: contribution, provider: Offer.create!(counter_contribution_params) }
    elsif contribution.offer?
      { receiver: Ask.create!(counter_contribution_params), provider: contribution }
    else
      # TODO: check if community resource type when it's added
      fail "Unhandled contribution type: #{contribution.type}"
    end
  end
```
Mostly just some inlining of methods that, for my eyes, makes the code a tad more readable.

Also note the `else` clause that fails fast instead of silently.

Oh, and i realize you didn't author this code and just moved it into the new interactor -- feel free to ignore this suggestion :)

app/interactions/create_match_for_contribution.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def counter_contribution_params
+    { person: person, service_area: contribution.service_area }
+  end
+
+  def person
+    match_with.person
```
exbinary, line 36: I wonder if it would be better to accept `person` in this interactor and let the controller pull the person out of user? I'm not sure which of these are the clearer semantics. Also not that important!

spec/interactions/add_person_details_from_claim_params_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    it "updates the person record with given details" do
+      person = user.person
+      expect(person).to have_attributes(
+        name: person.name,
+        preferred_contact_method: preferred_contact_method,
+        phone: contact_info
+      )
+    end
+  end
```
exbinary, line 35: I like the two contexts we're testing here but i'm not sure the assertions really guarantee the different behavior.

IIUC, there's nothing in the 'update' case ensuring that `user.person` wasn't replaced with a new object filled with the incoming params.

app/interactions/create_match_for_contribution.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class CreateMatchForContribution < ActiveInteraction::Base
```
exbinary, line 3: I'd consider naming this `MatchContribution` because in addition to creating a match, it also marks the existing contribution as 'matched'.

exbinary (approved):
Apologies for the much delayed review, @acherukuri ! 🐢

This is great! I like the refactorings, i'm curious your thoughts now that you've gone through the exercise.

I've added some comments/questions below. Feel free to ignore any of them as the code is already an improvement IMO.

🙏🏾💯

app/controllers/claims_controller.rb
```diff
@@ -14,22 +14,8 @@ def new
...
-                                                preferred_contact_method_id: claim_params[:preferred_contact_method_id],
-                                                contact_info: claim_params[:preferred_contact_info])
-      elsif current_person.present? && current_person.email.blank?
-        current_person.update!(email: claim_params[:preferred_contact_info])
-      end
-      Match.create_match_for_contribution!(contribution, current_user)
-      contribution.matched!
-    end
-    notify_peer_and_log_communication!(contribution)
+    claim_contribution!
```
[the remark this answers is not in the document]
acherukuri: Thanks for posting the link here. It's very informative. Honestly, I didn't know about that convention. I knew that bang methods mutate the receiver. In addition to that, I try to name methods with a bang to differentiate that they throw exceptions explicitly.

I can definitely change these to a non-bang version.

app/controllers/claims_controller.rb
```diff
@@ -39,20 +25,35 @@ def claim_params
     params.require(:claim).permit(:peer_alias, :preferred_contact_method_id, :preferred_contact_info, :message)
   end
 
-  def notify_peer_and_log_communication!(contribution)
-    peer_to_peer_email = PeerToPeerMatchMailer.peer_to_peer_email(
-      contribution,
-      peer_alias: claim_params[:peer_alias],
-      message: claim_params[:message],
+  def claim_contribution!
```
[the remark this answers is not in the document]
acherukuri: I have thought about that and definitely like the idea. I didn't really like the idea of having all the private methods (which don't belong here) defined in the controller layer. I will make those changes

app/interactions/create_match_for_contribution.rb
acherukuri: Renamed the file

app/interactions/create_match_for_contribution.rb
acherukuri: @exbinary I split the logic into multiple methods as the anchor method was getting bloated. But I think I broke it too much.

And I like the `else` clause to fail fast if the contribution type is unknown.

spec/interactions/add_person_details_from_claim_params_spec.rb
acherukuri: That's a good catch. I have added a test for that.

app/controllers/claims_controller.rb
```diff
@@ -14,46 +14,19 @@ def new
...
-
-    status = Messenger.new(peer_to_peer_email, 'peer_to_peer_email').deliver_now
-
-    CommunicationLog.log_email(
-      email: peer_to_peer_email,
-      delivery_status: status,
-      person: current_user.person,
-      initiator: current_user,
-    )
+    params.require(:claim).permit(:peer_alias, :email, :message)
```
exbinary, line 29: It's not necessary to use strong params here because the interactor defines exactly what's allowed and will ignore the rest.

app/controllers/claims_controller.rb
```diff
@@ -14,46 +14,19 @@ def new
...
-      end
-      Match.create_match_for_contribution!(contribution, current_user)
-      contribution.matched!
-    end
-    notify_peer_and_log_communication!(contribution)
+    ClaimContribution.run!(
+      contribution: params[:contribution_id],
+      claim_params: claim_params.to_h,
+      current_user: current_user
+    )
```
exbinary, line 22: I'd consider flattening the `claim_params` argument in the interactor into the three separate params, so that this code can be:
```suggestion
    ClaimContribution.run!(params.merge(
      contribution: params[:contribution_id],
      current_user: current_user
    ))
```

app/views/claims/new.html.erb
```diff
@@ -9,8 +9,7 @@
         
     <%= simple_form_for :claim, url: contribution_claims_path(contribution) do |f| %>
       <%= f.input :peer_alias, as: :string, label: false, placeholder: "Alias or handle (do not use your real name here)", required: true %>
-      <%= f.input :preferred_contact_method_id, collection: Array(ContactMethod.email), label: "Preferred contact method", required: true %>
-      <%= f.input :preferred_contact_info, as: :string, placeholder: "Add your phone number or email address", input_html: { value: preferred_contact_info }, required: true  %>
+      <%= f.input :email, as: :string, placeholder: "Add your phone number or email address", input_html: { value: preferred_contact_info }, required: true  %>
```
exbinary, line 12: The placeholder still mentions phone number. We could probably drop the placeholder TBH.
Also, is the reference to `.preferred_contact_info` still necessary?

exbinary (approved):
Looks great! Couple of minor comments.
Appreciate all your work on this! 🙏🏾 🚂

app/views/claims/new.html.erb
acherukuri: That's a good catch. I've cleaned this up and also made changes to the `new` action in controller.

maebeale (approved):
YESS!!!

PR #807, opened 2020-11-17
Allow anonymous users to generate CommunicationLogs

### LINKED ISSUE #805: 500 error when submitting an offer from a new person
Steps to recreate:

- Go to https://mutual-aid-nwmi.herokuapp.com/offers/new and fill out the form with a name and email address not already known to the system
- Fill out all the other required fields and submit the form

Expected behavior:

You will receive an indication your form was successfully submitted

Observed behavior:

HTTP 500 error. Seen in the server logs, this error is logged as:
```
Nov 12 16:51:07 mutual-aid-nwmi app/web.1 I, [2020-11-13T00:51:07.672086 #13]  INFO -- : [e70ed9c7-36b0-44fd-b056-f84803ad8620] Completed 500 Internal Server Error in 128ms (ActiveRecord: 54.0ms | Allocations: 19178)
Nov 12 16:51:07 mutual-aid-nwmi app/web.1 [e70ed9c7-36b0-44fd-b056-f84803ad8620] ActiveInteraction::InvalidInteractionError (User is required):
```

exbinary:
## Why
Fixes #805.

Integrating `EmailNewSubmission` into `Asks|OffersController` broke those forms for anonymous users. `CommunicationLog.log_email` allows a nil `initiator` object but the DB does not allow null in the corresponding `created_by_id` column.

## Pre-Merge Checklist

## How
First, the `EmailNewSubmission` interactor needed to allow a nil user (for when `current_user` is `nil`). But this by itself was not sufficient because the nil then causes `PG:NotNullViolation`.

`communication_logs.created_by_id` did have a default id set to 1, but this requires persistence code to omit the `created_by` field if it isn't provided; if a write is issued containing `created_by: nil`, rails will pass that on to the DB which then rejects it.

Moreover, i don't think it's wise to assume our 'system user' will always have an id of 1. If we really want CommunicationLog entries to have a system user as their default user id, we should explicitly set this in code.

Instead of implementing this additional logic, i've chosen here to drop the default value and changed the column to be nullable. As such, `null` can represent the same semantics of defaulting to the system user.

## Testing
- [x] Tests forthcoming

Would love your eyes on this @maebeale, particularly around the question of defaulting `created_by_id` to 1.

### DIFF
--- a/app/interactions/email_new_submission.rb
+++ b/app/interactions/email_new_submission.rb
@@ -2,7 +2,7 @@
 
 class EmailNewSubmission < ActiveInteraction::Base
   object :submission
-  object :user
+  object :user, default: nil
 
   def execute
     email = SubmissionMailer.new_submission_confirmation_email(submission)
--- a/app/models/communication_log.rb
+++ b/app/models/communication_log.rb
@@ -40,7 +40,7 @@ end
 #  subject            :string
 #  created_at         :datetime         not null
 #  updated_at         :datetime         not null
-#  created_by_id      :bigint           default(1), not null
+#  created_by_id      :bigint
 #  delivery_method_id :bigint           not null
 #  match_id           :bigint
 #  person_id          :bigint
--- a/db/migrate/20201113172851_allow_null_communication_logs_created_by_id.rb
+++ b/db/migrate/20201113172851_allow_null_communication_logs_created_by_id.rb
@@ -1,0 +1,5 @@
+class AllowNullCommunicationLogsCreatedById < ActiveRecord::Migration[6.0]
+  def change
+    change_column :communication_logs, :created_by_id, :bigint, default: nil, null: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_09_06_172102) do
+ActiveRecord::Schema.define(version: 2020_11_13_172851) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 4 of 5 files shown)


acherukuri (approved):

maebeale:
@acherukuri hey there! THANKS for working on this!!!

might we want to consider creating a user upon form submission?  
* a user that we don't tell the person we created, but we do it under the hood
* then the email confirmation could be sent later in the process to a real user, rather than trying to fake out the system's created_by_id data?

if not, i'm ok w removing the user requirement as you've done. just thinking it might be better practice to create the accounts rather than skip the user. (then, we have the additional bonus that after we have more user-facing-user-account pages, we can email all "users" to invite them to actually log in by resetting the password vs creating a new account AND setting the password. not sure everyone else would think this is a bonus, but... :) i do! lol.)

maebeale:
Regardless of whether we decide to silently create accounts or not, definitely agree w you @acherukuri that we shouldn't be defaulting to an id of 1, so glad that's been changed!

exbinary:
> @acherukuri hey there! THANKS for working on this!!!
> 
> might we want to consider creating a user upon form submission?
> 
> * a user that we don't tell the person we created, but we do it under the hood
> * then the email confirmation could be sent later in the process to a real user, rather than trying to fake out the system's created_by_id data?
> 
> if not, i'm ok w removing the user requirement as you've done. just thinking it might be better practice to create the accounts rather than skip the user. (then, we have the additional bonus that after we have more user-facing-user-account pages, we can email all "users" to invite them to actually log in by resetting the password vs creating a new account AND setting the password. not sure everyone else would think this is a bonus, but... :) i do! lol.)

Interesting line of thought! I find the approach of filling out a User under the hood to be appealing. Some questions:
* What would happen if someone fills out multiple anonymous requests? Would we have a way of reconciling/consolidating their separate user records when they eventually create an account?
* How would we then distinguish anonymous requests from those created by logged in users?

acherukuri:
> @acherukuri hey there! THANKS for working on this!!!
> 
> might we want to consider creating a user upon form submission?
> 
> * a user that we don't tell the person we created, but we do it under the hood
> * then the email confirmation could be sent later in the process to a real user, rather than trying to fake out the system's created_by_id data?
> 
> if not, i'm ok w removing the user requirement as you've done. just thinking it might be better practice to create the accounts rather than skip the user. (then, we have the additional bonus that after we have more user-facing-user-account pages, we can email all "users" to invite them to actually log in by resetting the password vs creating a new account AND setting the password. not sure everyone else would think this is a bonus, but... :) i do! lol.)

@maebeale It's @exbinary who worked on this 😀

PR #810, opened 2020-12-04
View contributions publicly when peer-to-peer is on

h-m-m:
## Why
 
This is one of the required steps of the peer-to-peer matchmaking workflow. This makes the list of available contributions generally public if the peer-to-peer setting is on.

## Next Steps

The next step after this will be to change what the "Respond" button does for people who aren't logged in. I'd prefer to keep that next step in a separate PR, but can add it on to here if people would rather keep them together.

Also, the permission logic here should be refactored out to a Pundit object once we're bringing those in. @maebeale talked about adding features later that allow users to mark their contributions as "for admin eyes only" or something similar. Implementation of those features would be a good time to refactor here.

### Pre-Merge Checklist

## What

- [x] moved the code that generates the urls for the response buttons out of the presenter and into the controller
- [x] updated controller before actions for authentication to include the new logic
- [x] refactored tests to describe the new behavior
- [ ] TODO: changed the `ContributionsController#respond` action because the view for that page currently has a lot of stuff in it that we don't want publicly-facing

### Accessibility

I expect this change will have no net effect on accessibility excepting any of the security concerns below.

### Security

There is a security risk here of inadvertently allowing information to be public.

- It is possible that someone will try submit some form of spam, advertising, or inappropriate promotion if info from the contribution forms is automatically made public like this
- It is very likely that some user will not understand what information on the forms they fill will or won't be made public
- It is possible take information that a user wants to and expects will stay private and make it public later. This can happen if an admin changes to p2p mode without notifying users
- an admin could also reveal info by changing the setting by accident

I feel like it's acceptable to continue with this work, understanding that the above security concerns are things we can address once we confirm mutual aid groups are happy with the basic functionality

### DIFF
--- a/app/controllers/contributions_controller.rb
+++ b/app/controllers/contributions_controller.rb
@@ -8,8 +8,8 @@ class ContributionsController < ApplicationController
 
   def index
     @filter_types = FilterTypeBlueprint.render([ContributionType, Category, ServiceArea, UrgencyLevel, ContactMethod])
-    filter = BrowseFilter.new(filter_params, self)
-    @contributions = ContributionBlueprint.render(filter.contributions, **filter.options)
+    filter = BrowseFilter.new(filter_params)
+    @contributions = ContributionBlueprint.render(filter.contributions, contribution_blueprint_options)
     respond_to do |format|
       format.html
       format.json { render inline: @contributions }
@@ -57,6 +57,13 @@ class ContributionsController < ApplicationController
   end
 
   private
+
+
+  def contribution_blueprint_options
+    options = { respond_path: ->(id) { respond_contribution_path(id)} }
+    options[:view_path] = ->(id) { contribution_path(id) } if SystemSetting.current_settings&.peer_to_peer?
+    options
+  end
 
   def filter_params
     return Hash.new unless allowed_params && allowed_params.to_h.any?
--- a/app/controllers/contributions_controller.rb
+++ b/app/controllers/contributions_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ContributionsController < ApplicationController
-  before_action :authenticate_user!, except: %i[combined_form respond thank_you]
+  before_action :authenticate_user!, unless: :peer_to_peer_mode?
   before_action :set_contribution, only: %i[respond triage]
 
   layout 'without_navbar', only: [:thank_you]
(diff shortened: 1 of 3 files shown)


app/controllers/contributions_controller.rb
```diff
@@ -58,6 +58,16 @@ def triage_update
 
   private
 
+  def peer_to_peer_mode?
+    @system_setting.peer_to_peer?
+  end
+
+  def contribution_blueprint_options
+    options = { respond_path: ->(id) { respond_contribution_path(id)} }
+    options[:view_path] = ->(id) { contribution_path(id) } if SystemSetting.current_settings&.peer_to_peer?
```
exbinary, line 67: Line 62 above references `@system_setting`. Can we do that here as well?

spec/requests/contributions_spec.rb
```diff
@@ -13,67 +13,90 @@
...
-      create(:listing)
-      get contributions_url
-      expect(response).to be_successful
-    end
+  ['when logged in', 'when logged out but p2p is enabled'].each do |scenario|
+    describe scenario do
+      before do
+        sign_in user unless scenario =~ /logged out/
+        allow_any_instance_of(SystemSetting).to receive(:peer_to_peer?).and_return(true) if scenario =~ /p2p is enabled/
+      end
```
exbinary, line 23: In general, i don't love repeating buckets of examples across multiple contexts.

There are of course circumstances when that's appropriate, eg for cross-cutting concerns. But in many cases, it can quickly become difficult to know which examples _really_ care about being tested in all contexts and which just happened to be lumped in out of convenience. This then can hinder future refactoring.

In this case, iirc the key thing to test between the three contexts is whether `/contributions` is accessible. I'd probably make that clear in a separate `describe 'access'` or similar block where that's the only thing being asserted. The examples below deal with details of the page which i don't think need to be re-run across contexts that are about authorization.

All that said, this is just my gut feeling -- i'm certainly fine with leaving it as is.

exbinary (approved):
Looks good. Agree the next chunk can be in a separate PR though maybe they need to be merged together?

Couple of comments in line, nothing majory.
💪🏾

maebeale (approved):
LGTM!

spec/requests/contributions_spec.rb
h-m-m: I've updated the tests as you've suggested

h-m-m:
Discussed with @maebeale and @svileshina, and I think we're okay with merging this in now, and then putting the "TODO: changed the ContributionsController#respond action because the view for that page currently has a lot of stuff in it that we don't want publicly-facing" part in a separate PR

As long as we don't cut a release before fixing that part of things, we're good

maebeale:
WOO HOOOOOO!

PR #816, opened 2021-01-08
focus outline rgba fix

svileshina:
## Why
rgba fix was not compiling


h-m-m (approved):
It also breaks on my machine and then works as soon as I pulled in this branch.

maebeale:
❤

PR #821, opened 2021-01-15
Show /contributions link in navbar when p2p enabled

h-m-m:
## Why
We want to show the /contributions link in navbar when p2p enabled because it's a key part of the p2p workflow

### Pre-Merge Checklist

## What
Screenshots of current code:

Logged out with P2P
Screen Shot 2021-01-14 at 7 48 16 PM

Logged out without P2P
Screen Shot 2021-01-14 at 7 45 02 PM

Logged in
Screen Shot 2021-01-14 at 7 40 09 PM

## How

We're adding a prop passed to Vue that indicates if P2P is on or not, and then letting Vue show or hide the button

### Testing
There are some JS tests since this is mostly a JS change

## Next Steps and Concerns

We talked some about rearranging these. Our concerns were:

- arranging buttons so items of higher priority or higher usefulness stand out more, which changes depending on context
- if someone logged in trying is to give instruction to someone not logged, it should be easy without the logged in person having to memorize which buttons are and are not the same for the logged out person
- having order be consistent and predictable
- understanding that most logged in users won't be spending much time if any looking at what a not logged in person sees

Because of all these concerns, we agreed to revisit how to balance them later

### Accessibility

We did not do accessibility testing for this. It seems unlikely to me that it would change the accessibility from the previous code

### Security

This may reveal the link URL in the JS to someone who is not logged in. That said, that URL will still have the Devise checks, so the additional insecurity is only removing obscurity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This changelog also serves to acknowledge the incredible people who've contribut
 * Unordered lists in user provided custom text (eg /about) now display with discs ('bullets') #777, #779
 * Allow custom text on landing page to override default text #769, #780
 * Add Rubocop #793
+* Show the "Contributions" link in the navbar for P2P users even if they aren't logged in
 
 ## [0.3.1] - 2020-10-10
 ### Enhancement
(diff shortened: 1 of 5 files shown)


exbinary:
Mob reviewed!

PR #823, opened 2021-01-18
Restore ability to hit /thank_you page w/o logging in

exbinary:
Was mistakenly dropped in e301ccd9, but is necessary even in dispatch mode so that guests can create contributions.

https://user-images.githubusercontent.com/8330/104955888-f7aad080-5998-11eb-9b62-1a3986e3c4d0.mov


exbinary:
Reviewed in person with @h-m-m .

Issue #819, opened 2021-01-14
Finish up P2P MVP


exbinary:
When P2P mode is enabled:
- [x] Contribution button on NavBar should be accessible w/o login
- [x] Respond buttons on /contributions should only be visible to admins (#824)
- [x] Fix bug with `ClaimContribution` interactor missing params (#832)
- [x] Check if claiming sends an email
- [x] Remove person name in map view #847 
- [x] Re-enable registration #840 
- [x] ~Make email required on ask/offer forms (until sms is ready)~
- [x] Prevent claiming email-less contribution. #831
- [x] Redesign From section of Claim form, drop ability to override email, add disclaimer text no how to change your account email if you don't want others to see it. #855 
- [x]  View button on List view and Map view

Nice to have: 
- [ ] Implement `description` from #310.
- [ ] Add note in ContactMethods re looking up fontawesome icons

And then:
- [x] Cut a release!

exbinary:
Done! w00t!

PR #824, opened 2021-01-18
Hide the "respond" UI items when not logged in

h-m-m:
## Why
This addresses one of the items in #819

We'll probably want more logic involved later. For now, it's okay hide all "respond" UI items when not logged in. The only time they should show up for users not logged in is when the system is in peer-to-peer mode

### Pre-Merge Checklist

## What

Also, I added some tests, pulling some test data from my dev environment

When not logged in, viewing tiles:
Screen Shot 2021-01-18 at 2 01 46 PM

When not logged in, viewing the list:

Screen Shot 2021-01-18 at 2 01 40 PM

When logged in viewing tiles:
Screen Shot 2021-01-18 at 2 02 04 PM

When logged in viewing the list:
Screen Shot 2021-01-18 at 2 01 16 PM

## How
The two key changes are 

- a line in the controller that makes the respond link conditional
- a line in the list view that hides the "Respond" header of no controllers have data for it

### Accessibility

This should not change accessibility as it makes an existing UI element entirely hidden conditionally

### Security

This is pretty reasonable, security-wise. The decision to expose the information is being made server-side, so there's no way a user could spy on the info only from the client. That, and the URLs that are being exposed have authorization controls around them, so there's not much risk if they were exposed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This changelog also serves to acknowledge the incredible people who've contribut
 * Allow custom text on landing page to override default text #769, #780
 * Add Rubocop #793
 * Show the "Contributions" link in the navbar for P2P users even if they aren't logged in
+* Hide the respond button on contributions when nobody is logged in. (We'll add more functionality later for the peer-to-peer scneario) #819
 
 ## [0.3.1] - 2020-10-10
 ### Enhancement
--- a/spec/javascript/pages/browse/ListBrowser.spec.js
+++ b/spec/javascript/pages/browse/ListBrowser.spec.js
@@ -4,21 +4,33 @@ import ListBrowser from 'pages/browse/ListBrowser'
 import testData  from '../../../../lib/contributions.json'
 
 describe('ListBrowser', () => {
-  it('shows the respond column with typical data', function () {
-    const wrapper = shallowMount(ListBrowser, {
+  def('contributions', testData)
+
+  def('wrapper', () => {
+    return shallowMount(ListBrowser, {
       localVue: configure(createLocalVue()),
-      propsData: testData
+      propsData: $contributions,
     })
-    var respondHeaders = wrapper.findAll('th').filter( header => header.text() == 'Respond')
-    assert.equal(1, respondHeaders.length)
   })
-  it('hides the respond if there is no response urls', function() {
-    var responselessContributions = testData.contributions.map( contribution => { contribution.respond_path = ''; return contribution})
-    const wrapper = shallowMount(ListBrowser, {
-      localVue: configure(createLocalVue()),
-      propsData: {responselessContributions}
+
+  def('tableHeaders', () => $wrapper.findAll('th').wrappers)
+
+  describe('respond column', () => {
+    it('shows the column with typical data', function () {
+      assert.isTrue($tableHeaders.some(header => header.text() == 'Respond'))
     })
-    var respondHeaders = wrapper.findAll('th').filter( header => header.text() == 'Respond')
-    assert.equal(0, respondHeaders.length)
+
+    describe('when contributions do not come with response urls', () => {
+      def('contributions', () => {
+        testData.contributions.map(contribution => {
+          contribution.respond_path = ''
+          return contribution
+        })
+      })
+
+      it('hides the respond button', function() {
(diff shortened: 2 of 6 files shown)


h-m-m:
Please critique my Vue and JS testing, as it's been since whenever I last touched JS on this project that I wrote anything in Vue or tried to test JS. I'm particularly interested in cleaner ways of expressing the tests due to not remembering all the methods that are available

spec/javascript/pages/browse/ListBrowser.spec.js
```diff
@@ -1,16 +1,24 @@
...
+    assert.equal(1, respondHeaders.length)
+  })
+  it('hides the respond if there is no response urls', function() {
+    var responselessContributions = testData.contributions.map( contribution => { contribution.respond_path = ''; return contribution})
+    const wrapper = shallowMount(ListBrowser, {
+      localVue: configure(createLocalVue()),
+      propsData: {responselessContributions}
+    })
+    var respondHeaders = wrapper.findAll('th').filter( header => header.text() == 'Respond')
+    assert.equal(0, respondHeaders.length)
```
exbinary, line 22: Refactored a bit to showcase some rspec-like techniques you could consider:
```suggestion
  def('contributions', testData)

  def('wrapper', () => {
    return shallowMount(ListBrowser, {
      localVue: configure(createLocalVue()),
      propsData: $contributions,
    })
  })

  def('tableHeaders', () => $wrapper.findAll('th').wrappers)

  describe('respond column', () => {
    it('shows the column with typical data', function () {
      assert.isTrue($tableHeaders.some(header => header.text() == 'Respond'))
    })

    describe('when contributions do not come with response urls', () => {
      def('contributions', () => {
        testData.contributions.map(contribution => {
          contribution.respond_path = ''
          return contribution
        })
      })

      it('hides the respond button', function() {
        assert.isFalse($tableHeaders.some(header => header.text() == 'Respond'))
      })
```
Things to note:
* Use of `def` which is a like `let` in rspec. Helpful to extract common pieces (eg wrapper creation logic), thereby making the difference between examples more distinct.
* Use of `some` array method instead of `find` + `length`.

exbinary (approved):
LGTM! Made some vue-testing suggestions as requested, though nothing that necessarily needs changed.

Not specific to this PR but i'm not a fan of using big chunks of json fixture data in specs. Among other issues, they often change drastically between commits and it's very difficult to know what part of the diff were germane to the code/spec changes and what is coincidental.

h-m-m:
> LGTM! Made some vue-testing suggestions as requested, though nothing that necessarily needs changed.

Thanks!
 
> Not specific to this PR but i'm not a fan of using big chunks of json fixture data in specs. Among other issues, they often change drastically between commits and it's very difficult to know what part of the diff were germane to the code/spec changes and what is coincidental.

I agree with those negatives. I'd still some way to verify that the data that we expect Rails to emit is then consumed properly by by the JS. Do you have any suggestions for that?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This changelog also serves to acknowledge the incredible people who've contribut
 * Allow custom text on landing page to override default text #769, #780
 * Add Rubocop #793
 * Show the "Contributions" link in the navbar for P2P users even if they aren't logged in
+* Hide the respond button on contributions when nobody is logged in. (We'll add more functionality later for the peer-to-peer scneario) #819
```
maebeale, line 19: ❤️

spec/javascript/pages/browse/ListBrowser.spec.js
```diff
@@ -1,16 +1,36 @@
-import {assert} from 'chai'
```
maebeale, line 1: why isn't assert needed from chai anymore?

maebeale (approved):
Love it!

spec/javascript/pages/browse/ListBrowser.spec.js
h-m-m: I think that's because our `setup.js` file brings it in, and that setup file is explicitly included in https://github.com/rubyforgood/mutual-aid/blob/8086ba858d0f760088ed0785b4f68e706cff3121/package.json#L50

PR #832, opened 2021-01-19
Fix missing params error when claiming

exbinary:
### Why
I introduced a regression in 7fa64948 by not accounting for params being nested under `params[:claim]` 🤕 

### Pre-Merge Checklist

### Testing, Outstanding Concerns
Really speaks to the need for integration tests!


maebeale (approved):
! :)

PR #833, opened 2021-01-20
Drop pundit rspec shared contexts

exbinary:
### Why
Inlining these snippets creates a smidge of duplication but i believe this is a worthy tradeoff for simplicity, at least for now. In particular, this eliminates the coupling amongst and between shared contexts and specs via the shared `acting_user` let; such implicit coupling can be surprising, hard to recognize, and cause hurdles when refactoring.

### Pre-Merge Checklist

### What
- [x] Remove `shared_contexts/authorization`
- [x] Refactor `person_policy_spec` and `user_policy_spec` to use regular `let`s instead
- [x] Change `requests/user_spec` and `requests/person_spec` to explicitly create and log_in a sysadmin user
- [x] Drop related references in `rails_helper`

### Next Steps
Sets us up to rename the `acting_user` accessor in `ApplicationPolicy` if we choose to.


svileshina (approved):
🥳  looks good!

PR #834, opened 2021-01-20
Rename acting_user accessor in policies

exbinary:
### Why
Definite tradeoffs in both directions but after a group discussion, felt that new engineers having to investigate/learn another term wasn't worth the benefits of the explicit qualification.

### Pre-Merge Checklist
- [ ] Talk to @bhaibel 

### What
1. Dropping back to `user` -- the term used in pundit docs/examples with the understanding that any _other_ users in play can be qualified as needed.
1. Rename a couple of related private methodes in `UserPolicy` and `PersonPolicy`
1. Also renames `original_scope` to `scope`. Doesn't feel like the qualification is necessary so falling back to the
term used in pundit docs/examples.

Note: this changeset builds on #833 which first drops some spec code that uses the same term.

We discussed that we'd love @bhaibel  to weigh in on this since she likely considered these tradeoffs. We'd ideally not merge these suggested changes until we have a chance to talk to her.

### DIFF
--- a/app/policies/person_policy.rb
+++ b/app/policies/person_policy.rb
@@ -4,8 +4,8 @@ class PersonPolicy < ApplicationPolicy
       case
       when can_admin?
         original_scope.all
-      when acting_user.present?
-        original_scope.where(user_id: acting_user.id)
+      when user.present?
+        original_scope.where(user_id: user.id)
       else
         original_scope.none
       end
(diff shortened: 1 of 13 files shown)


maebeale (approved):

app/policies/person_policy.rb
```diff
@@ -3,42 +3,38 @@ class Scope < Scope
...
+        scope.none
       end
     end
   end
 
   def read?
-    person_attached_to_acting_user? ||
-      sys_admin? ||
-      admin?
+    own_person? || sys_admin? || admin?
```
maebeale, line 16: `own_person` takes me a minute to translate, but, I don't have a different suggestion atm that doesn't have a namespace issue. :)

maebeale (approved):

maebeale (approved):

exbinary:
@maebeale , i brought this up to date with latest main. Please give it another look when you get a chance.

PR #835, opened 2021-01-21
Solution to secure admin-only resources

exbinary:
### Why
Before we open up user registration, we have to secure any pages that should only be accessible to admins (sysadmins). This actually turns out to be the vast majority of our controllers, so ideally we can solve this w/o a lot of drudge work.

### Pre-Merge Checklist

### What
- [x] Add code to `AdminController` that ensures current_user is an admin || sysadmin
- [x] Show sample implementation with one admin-only controller (`ContactMethodsController`)

### How
* With `pundit`, we would usually `authorize` each action independently and create policy classes for each resource.
* For admin-only resources, this seems like overkill: these controllers and actions all require exactly the same check.
* Inheriting this logic from a base controller in the form of a `before_action` seems like a simpler solution.
* Currently, `AdminController` contains a bunch of non-restful actions. I recommend we extract them out into their own controllers so `AdminController` can become an abstract base class.

### Testing
Added a spec specifically around the simple auth introduced in `AdminController`. It uses `ContactMethodsController` as a sample admin-only resource.

### Next Steps
- [ ] Extract restful controllers out of the existing actions in `AdminController` since we don't want all subclasses to inherit this extraneous logic (#836).
- [ ] Make other admin-only controllers subclasses of `AdminController` (#837).

?

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1,9 +1,8 @@
 # frozen_string_literal: true
 
 class AdminController < ApplicationController
-  include NotUsingPunditYet
-
-  before_action :authenticate_user!, except: %i[new create]
+  before_action :authenticate_user!
+  before_action :ensure_authorized_as_admin
   before_action :set_system_settings, only: [:glossary_edit, :glossary_index, :glossary_update]
 
   def landing_page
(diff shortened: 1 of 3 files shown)


maebeale:
I think someday we should have two levels of admin, but agree this should be the path until then

maebeale:
Support removing NON-RESTful actions!

exbinary:
Ready for review. H/T @h-m-m for his help simplifying the auth implementation 🍠 .

h-m-m (approved):
I'm pretty happy about where we landed with this

PR #837, opened 2021-01-23
Add simple auth to all admin-only controllers

exbinary:
### Why
Gives all admin-only controllers rudimentary authorization by inheriting from `AdminController`.

See #835 for background.

### Pre-Merge Checklist

### Testing
This is part of the work to re-open user registration.
Once all that work is ready, we'll need to do some extensive manual regression testing.

### Specific review requests
Most importantly, please check whether all these controllers are actually only for admin use.

### Next Steps
?

?

### DIFF
--- a/app/models/listing.rb
+++ b/app/models/listing.rb
@@ -67,7 +67,7 @@ class Listing < ApplicationRecord
     elsif matches_as_provider.any?
       status = matches_as_provider.map{|m| m.completed?}.any? ? 'completed' : 'matched'
     end
-    update_attributes(state: status)
+    update(state: status)
     status
   end
 
--- a/spec/requests/listings_spec.rb
+++ b/spec/requests/listings_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe '/listings', type: :request do
     location_attributes: { zip: '12e45' },
   }}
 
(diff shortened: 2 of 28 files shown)


app/models/listing.rb
```diff
@@ -67,7 +67,7 @@ def status
     elsif matches_as_provider.any?
       status = matches_as_provider.map{|m| m.completed?}.any? ? 'completed' : 'matched'
     end
-    update_attributes(state: status)
+    update(state: status)
```
exbinary, line 70: Gets rid of an annoying deprecation warning :)

spec/requests/listings_spec.rb
```diff
@@ -13,7 +13,7 @@
     location_attributes: { zip: '12e45' },
   }}
 
-  before { sign_in create(:user) }
+  before { sign_in create(:user, :admin) }
```
exbinary, line 16: I believe `ListingController` is only currently used by admins?

exbinary:
Pair reviewed with @maebeale .

maebeale (approved):
Perfecto! Some of these are likely to need to come out of this category, but until we have the relevant functionality, all of these should be restricted to admin/sys_admin only. Tks!

PR #844, opened 2021-01-28
Introduce context

exbinary:
### Why
This changeset was motivated by the proliferation of inherited @variables in many of our controllers, eg `@system_setting`.

It's bad practice to reference superclass instance variables because it creates very tight coupling up and down the inheritance chain. These variables essentially become globals.

In rails, this problem is usually solved in one of several ways. Taking `@system_setting` as an example:
* Wrap variables in attr_readers, eg `attr_reader :system_setting`. This is an improvement but still maintains coupling between super and sup classes.
* Mix in a module containing an accessor. This is an improvement because it doesn't force accessors onto other classes in the inheritance chain that don't need it. There are still a couple of downsides:
  * There are usually a number of accessors in play; authors have to discover and bring in multiple modules.
  * Because accessors are mixed directly into the controller, the coupling is still tight; there isn't a distinct 'seam' between the controller and the dependency, so we have to mock/override accessors. Again, if there are several accessors in play, this can get cumbersome.
* Dropping the accessor and getting the underlying data directly, usually via a class method, eg `SystemSetting.current_settings`. Eliminates the sub/super coupling, replacing it with direct coupling; the coupling shifts from 'vertical' to 'horizontal'. This makes the seam much clearer but still suffers violates the Dependency Inversion principle and can also lead to duplication.

Further, all the solutions above only help at the controller layer. We often need to access some of these variables in other layers such as Interactors or Serializers. A good example of this is needing access to `current_user` outside of controllers.

This changeset outlines an alternative that i believe minimizes all these concerns.

### What and How
  1. Introduces `Context`, a thin wrapper providing access to 'globals' such as `system_settings` and `current_user`. This serves as a single point of access to any 'context' we might need.
  1. `Context` can be initialized with data or filled out later via attr_writers.
  1. It can also lazily retrieve and memoize data (see Context#system_settings).
  1. This new class turns out to be a practical and semantic superset of `UserContext`, so they've been consolidated.
  1. Shows a simple example of usage in `AnnouncementsController`

### Testing
- [x] `app/models/context_spec.rb`

### Next Steps
- [x] Refactor `AnnouncementsController` to drop `@admin_status` in favor of `context.can_admin?` #845 
- [ ] ?

?

### Meta
Very open to discussion and feedback. I realize this approach bucks the typicals 'rails way'.

### Pre-Merge Checklist

### DIFF
--- a/spec/models/context_spec.rb
+++ b/spec/models/context_spec.rb
@@ -36,4 +36,19 @@ RSpec.describe Context do
       context.system_settings
     end
   end
+
+  describe 'to_h' do
+    let(:user) { double 'User' }
+    let(:system_settings) { double 'SystemSettings' }
+    let(:context) { Context.new user: user, system_settings: system_settings }
+
+    def method_with_keyword_args positional_arg, non_context_keyword_arg:, user:, system_settings:
+      [positional_arg, non_context_keyword_arg, user, system_settings]
+    end
+
+    it 'can be converted to a hash and passed as keyword args' do
+      expect(method_with_keyword_args 'positional', non_context_keyword_arg: 'non_context', **context.to_h).
+        to eq ['positional', 'non_context', user, system_settings]
+    end
+  end
 end
--- a/spec/models/context_spec.rb
+++ b/spec/models/context_spec.rb
@@ -1,0 +1,38 @@
+require 'spec_helper'
+
+RSpec.describe Context do
+  describe 'initialization' do
+    it 'allows keyword init' do
+      context = Context.new user: double(:user)
+      expect(context.user).to be_present
+    end
+
+    it 'allows args to be omitted' do
+      context = Context.new
+      expect(context.user).to be_nil
+    end
+  end
+
+  describe 'lazy, memoized, overridable readers' do
+    let(:current_settings) { double :current_settings }
+    let(:system_setting) { class_double('SystemSetting').as_stubbed_const }
+    let(:context) { Context.new }
+
+    it 'is lazy' do
+      expect(system_setting).to receive(:current_settings) { current_settings }
+      expect(context.system_settings).to be current_settings
+    end
+
+    it 'is memoized' do
+      expect(system_setting).to receive(:current_settings).once { current_settings }
+      context.system_settings
+      context.system_settings
+    end
+
+    it 'can be overriden' do
+      expect(system_setting).to receive(:current_settings).never
+      context.system_settings = double(:override)
+      context.system_settings
+    end
+  end
+end
--- a/spec/models/context_spec.rb
+++ b/spec/models/context_spec.rb
@@ -13,23 +13,24 @@ RSpec.describe Context do
     end
   end
 
-  describe 'lazy, memoized, overridable readers' do
+  describe 'custom reader methods' do
     let(:current_settings) { double :current_settings }
     let(:system_setting) { class_double('SystemSetting').as_stubbed_const }
     let(:context) { Context.new }
 
-    it 'is lazy' do
-      expect(system_setting).to receive(:current_settings) { current_settings }
+    example 'can be lazy' do
+      expect(context[:system_settings]).to be_nil # not provided during initialization
+      expect(system_setting).to receive(:current_settings).and_return current_settings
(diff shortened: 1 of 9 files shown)


exbinary:
H/T @h-m-m for bringing [`CurrentAttributes`](https://devdocs.io/rails~6.0/activesupport/currentattributes) to my awareness: TIL!

While it is certainly solving the same problem i fundamentally disagree with its design (which actively violates the dependency inversion principle). That said, i'm open to using it instead if folks see value in sticking to rails conventions.

exbinary:
Realized this PR was doing too much and the policy related changes were distracting from the primary design suggestion. Extracted #849 to hold the `?admin=false` related changes.

spec/models/context_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  describe 'lazy, memoized, overridable readers' do
+    let(:current_settings) { double :current_settings }
+    let(:system_setting) { class_double('SystemSetting').as_stubbed_const }
+    let(:context) { Context.new }
+
+    it 'is lazy' do
+      expect(system_setting).to receive(:current_settings) { current_settings }
+      expect(context.system_settings).to be current_settings
```
h-m-m, line 23: Maybe because I'm a little rusty with the RSpec matchers here, I'm not sure how this definitively tests the laziness. A non-lazy implementation wouldn't pass this test?

exbinary: Good point! I think i was making the mental leap that since we didn't pass a `SystemSetting` into the `Context` initializer, then verifying the getter returns the current one _infers_ a lazy lookup. But that's all just in my head :)

I'll change this to make the expectation and method of verification clearer.

exbinary: Attempted to clarify in 1bac57d.

maebeale (approved):

exbinary:
As discussed, i'm going to merge this but wanted to document an answer to the question @h-m-m raised in person: why do i prefer this approach to the rails [CurrentAttributes](https://devdocs.io/rails~6.0/activesupport/currentattributes)?

They both solve the same problem.
In short, `CurrentAttributes` violates the Dependency Inversion Principle.

Here's an example contrasting each solution when an interactor needs `SystemSetting`s (as an example of a 'global').

#### Current
```ruby
# interactor
class EmailNewSubmission < ActiveInteraction::Base
  object :submission

  def execute
    SubmissionMailer.new_submission_confirmation_email(submission, Current.system_settings)

# controller
class OffersController < PublicController
  def create
    EmailNewSubmission.run! submission: submission
```
(Unrelated code ommitted)

Here you can see system settings being retrieved from within the interactor via `Current.system_settings`.

Note this dependency is not visible when looking at the controller, nor when scanning the params defined at the top of the interactor. The dependency is buried within the execute method.

#### Context
```ruby
# interactor
class EmailNewSubmission < ActiveInteraction::Base
  object :submission
  object :system_setting

  def execute
    SubmissionMailer.new_submission_confirmation_email(submission, system_setting)

# controller
class OffersController < PublicController
  def create
    EmailNewSubmission.run! submission: submission, system_setting: context.system_settings
```
In contrast, here the dependency is explicit and obvious.

The downside tradeoff is that if a dependency needs to be passed through multiple layers of code, this can be cumbersome. Imagine for example that the interactor calls another interactor which then calls a mailer which needs system settings. All of these intermediate objects would then need to pass-through a param, whereas with the `Current` solution, the mailer could directly retrieve system settings.

IMO, the 'pain' of having to pass that param through is actually very useful; it acts as a 'smell', encouraging us to rethink our design. There are certain cases, such as some cross-cutting concerns, where there aren't clear alternatives but these are the exception to the general rule.

A secondary argument is that `CurrentAttributes` uses 'magic' to make sure `Current.system_settings` is isolated across requests although it's accessed through a class method (which in plain ruby would usually access global data). Where possible, i prefer to avoid such magic that goes against the grain of plain ruby.

`context` on the other hand is a just an accessor around an instance variable lazily initialized in `ApplicationController` which already guarantees request isolation (because each request get's a new instance of the corresponding controller).

PR #846, opened 2021-01-30
Disable claim button when p2p and no email related to contribution

### LINKED ISSUE #831: Prevent p2p user from claiming a contribution that does not have an email.

AngelaGuardia:
### Why
closes #831 

This is a temporary feature added because we have yet to set up a way to allow users to communicate directly with each other (p2p) over SMS therefore we can only allow users to claim 'asks' that have emails related to them.

### Pre-Merge Checklist

### What
Add logic to the contributions show view to disable the claim button when the following conditions are met:
- The system is in peer to peer mode
- The contribution's owner did not provide an email
Screen Shot 2021-01-30 at 12 18 09 PM

### How
- Added a `@peer_to_peer` instance variable to the contributions_controller show action that holds a boolean that represents whether the system is in p2p mode
- Used the `@peer_to_peer1 variable as part of a conditional to disable the claim button
- The second part of the condition is whether or not the owner of the contribution has an email

### Testing
I could use some help adding tests for this feature. The only way I know how to test these front-end components is using Capybara but it looks like here you're using "interactions"?

I tested it manually as follows:
- Changed the system settings to p2p
- Filtered contributions by excluding those with email
- Asserted that the claim button was disabled

- Changed the system settings to fully moderated
- Filtered contributions by excluding those with email
- Accessed a contribution's show page (had to access via URL since this mode had no "view" button in the contributions index page)
- Asserted that the claim button was enabled

### Next Steps
- This feature will be removed once we have SMS set up

- Left some questions in the PR review

### Accessibility
- I believe this view should have a flash message saying that contributions without email cannot be claimed but I did not see that locally
- Maybe we could add some permanent disclaimer text near the disabled button?
### Meta
- Had some issues finding the username & password pairs to manually test this! are they documented somewhere? I ended up using 
- `mutualaidtesting@example.com` `testing123`
- `alene_donnelly@herman.biz` `password-anne-rutherford` (found this after digging in the seeds for a while)


app/controllers/contributions_controller.rb
```diff
@@ -20,6 +20,7 @@ def index
 
   def show
     contribution = Listing.find(params[:id])
+    @peer_to_peer = peer_to_peer_mode?
```
AngelaGuardia, line 23: I tried to pass this in as a local variable (like contribution is passed on line 28) to follow the pattern here more closely but couldn't. It was not accessible from the view. So I settled with this "@" variable. Is there a preference for one or the other?

leahriffell:
Nice, Angela! I like how clean this is as an interim solution. I do think that adding some explanatory text near the button when disabled would be helpful. Maybe something like "An ask without an email address cannot be claimed", "An ask without an email address cannot be claimed in peer-to-peer mode", etc.

app/views/contributions/show.html.erb
```diff
@@ -7,8 +7,8 @@
     <p class="subtitle"><%= contribution.description %></p>
     <% if contribution.matched? %>
       <i class="far fa-check-circle has-text-primary is-size-5 has-text-weight-semibold"><span class="ml-1">Claimed</span></i>
-    <% else %>
-      <%= button_to "Claim This #{contribution.type}", new_contribution_claim_path(contribution), class: "button is-primary", method: :get %>
+    <% else  %>
+      <%= button_to "Claim This #{contribution.type}", new_contribution_claim_path(contribution), class: "button is-primary", disabled: (@peer_to_peer && !contribution.person.email), method: :get %>
```
h-m-m, line 11: I have a preference for not having compound logic like this in the view. Pulling it out of the view makes it easier to write a test around it

Some suggestions on alternative places to put this:
* a helper method
* a method on contribution
* a presenter object, a new plan Ruby class or a `SimpleDelegator` class, that accepts a contribution and adds methods to it like `publicly_claimable?`

I'm happy to go into my thoughts on this in more detail if you'd like

Gemfile.lock
```diff
@@ -162,76 +162,77 @@ GEM
...
-      actionmailer (= 6.0.2.2)
-      actionpack (= 6.0.2.2)
-      actiontext (= 6.0.2.2)
-      actionview (= 6.0.2.2)
-      activejob (= 6.0.2.2)
-      activemodel (= 6.0.2.2)
-      activerecord (= 6.0.2.2)
-      activestorage (= 6.0.2.2)
-      activesupport (= 6.0.2.2)
+    rails (6.0.3.4)
```
h-m-m, line 165: I'm generally fine with bumping the Rails minor version. I think it deserves its own separate PR if that's what you'd like to do

h-m-m (changes requested):

app/views/contributions/show.html.erb
```diff
@@ -7,8 +7,8 @@
     <p class="subtitle"><%= contribution.description %></p>
     <% if contribution.matched? %>
       <i class="far fa-check-circle has-text-primary is-size-5 has-text-weight-semibold"><span class="ml-1">Claimed</span></i>
-    <% else %>
-      <%= button_to "Claim This #{contribution.type}", new_contribution_claim_path(contribution), class: "button is-primary", method: :get %>
+    <% else  %>
+      <%= button_to "Claim This #{contribution.type}", new_contribution_claim_path(contribution), class: "button is-primary", disabled: not_claimable?(contribution), method: :get %>
```
AngelaGuardia, line 11: @h-m-m Is using the `not_claimable?` helper method like this more along the lines of what you were thinking?

I was discussing this with @leahriffell and we were thinking that we wanted to put the helper method in the contribution model instead that way we could do a unit test on it but I'm not sure how the current `contribution.rb` file works with the `TYPES = [Ask, Offer]`

Gemfile.lock
AngelaGuardia: I didn't realize I had bumped this up. What would be the best way to avoid this in the future? Change the rails version that I'm running locally?

app/views/contributions/show.html.erb
h-m-m: I definitely think this is an improvement. I checked with @exbinary, who thinks we're going to want to make further changes to the logic soon. He may have some further suggestions, but whatever that plan is, pulling the logic out like this will make it easier for us to get there

The only remaining question I have, aside from testing concerns, is if it's better to disable the button or hide it. Or if we disable it, should there be some other message to the user?

I don't know what's the right thing here from a UX perspective, so I'm curious what you think

Gemfile.lock
h-m-m: The Gemfile.lock and yarn.lock files exist so that we can make sure we're all using the same packages even if our machines aren't configured exactly the same. I imagine some command you ran updated them. There are several commands that could do this, so it's hard to say exactly and may not be worth your time to track it down right now

See if you can revert the changes to these lock files and still get the application to run in your local environment. You should be able to do it using only instructions listed in https://github.com/rubyforgood/mutual-aid/blob/main/doc/setup.md without any changes to these files

 If that doesn't work for you, let me know and we can troubleshoot the problems you're having tomorrow

app/views/contributions/show.html.erb
leahriffell: I would think that disabling the button _and_ adding an explanatory message would be a less confusing experience than the button disappearing completely. Maybe something like "An ask without an email address cannot be claimed", "An ask without an email address cannot be claimed in peer-to-peer mode", etc.

AngelaGuardia: I agree with Leah. I like the idea of disabling it with a message better than hiding it. When we were looking at this together last week there was a flash message that explained the situation but I don't see it anymore. Regardless, I think the message should be permanent and near the button for better UX.

app/views/contributions/show.html.erb
```diff
@@ -7,8 +7,13 @@
...
       <i class="far fa-check-circle has-text-primary is-size-5 has-text-weight-semibold"><span class="ml-1">Claimed</span></i>
-    <% else %>
-      <%= button_to "Claim This #{contribution.type}", new_contribution_claim_path(contribution), class: "button is-primary", method: :get %>
+    <% else  %>
+      <% if contribution.has_email? %>
+        <%= button_to "Claim This #{contribution.type}", new_contribution_claim_path(contribution), class: "button is-primary", method: :get %>
+      <% else %>
+        <%= button_to "Claim This #{contribution.type}", new_contribution_claim_path(contribution), class: "button is-primary", disabled: true, method: :get %>
+        <p class="is-6 is-italic">An ask without an email address cannot be claimed in peer-to-peer mode</p>
+      <% end %>
```
leahriffell, line 16: What do y'all think of this change to add explanatory text next to the button when disabled? Is there a cleaner way to do this without another conditional? 

Also, I added `is-6` and `is-italic` classes because I saw those on the bulma docs. I looked through our styling and didn't see anything related to explanatory-type formatting.

exbinary: Looks good @leahriffell . I don't recall anything in bulma specifically for this kind of thing; i think a `p` with those classes is very reasonable.

app/views/contributions/show.html.erb
```diff
@@ -7,8 +7,13 @@
     <p class="subtitle"><%= contribution.description %></p>
     <% if contribution.matched? %>
       <i class="far fa-check-circle has-text-primary is-size-5 has-text-weight-semibold"><span class="ml-1">Claimed</span></i>
-    <% else %>
-      <%= button_to "Claim This #{contribution.type}", new_contribution_claim_path(contribution), class: "button is-primary", method: :get %>
+    <% else  %>
+      <% if contribution.has_email? %>
+        <%= button_to "Claim This #{contribution.type}", new_contribution_claim_path(contribution), class: "button is-primary", method: :get %>
+      <% else %>
+        <%= button_to "Claim This #{contribution.type}", new_contribution_claim_path(contribution), class: "button is-primary", disabled: true, method: :get %>
```
exbinary, line 14: Very minor thought: you could simplify line 14 to something like:
```suggestion
        <button class="button is-primary" disabled="true">Claim This <%= contribution.type %></button>
```
because the disabled button can omit all the extra details needed (and rendered) by `button_to`.

.ruby-version
```diff
@@ -1 +1 @@
-ruby-2.7.0
+2.7.0
```
exbinary, line 1: I'm curious what triggered this change?

app/views/contributions/show.html.erb
leahriffell: Great catch, thank you @exbinary!

.ruby-version
ETBassist: @rodri0315 and I think we've fixed this 👍

exbinary:
Thanks all!

exbinary (approved):

PR #847, opened 2021-01-31
Remove name in map view

AngelaGuardia:
### Why
Improving user privacy by removing person name from the contributions map view
Addresses one item in #819 

### Pre-Merge Checklist

### What
-  prevent person name from showing on map view

Before
Screen Shot 2021-02-03 at 1 54 18 PM

After
Screen Shot 2021-02-03 at 1 59 40 PM

### How
- Removed `(person.name)` from `listing.rb` `name` method

### Testing
- No tests were added for this. Tested visually.

- The current solution may affect other views in the app where the person name is desired
- Instead of modifying the `name` method, I wanted to create a new method in `listing.rb` called `private_name` and call that on line 63 of `MapBrowser.vue`. That way I could be certain that this change would not affect other areas of the app. I wasn't able to make that work but it may be a better solution.

### DIFF
--- a/app/models/listing.rb
+++ b/app/models/listing.rb
@@ -51,7 +51,7 @@ class Listing < ApplicationRecord
   end
 
   def name
-    type + ": #{all_tags_to_s} (#{person.name})"
+    type + ": #{all_tags_to_s}"
   end
 
   def name_and_match_history
--- a/spec/models/listing_spec.rb
+++ b/spec/models/listing_spec.rb
@@ -11,6 +11,13 @@ RSpec.describe Listing, type: :model do
 
       expect(Listing.tagged_with('cash')).to match_array([listing])
     end
+
+    example 'name method' do
+      listing.tag_list.add('grocery shopping', 'childcare')
+      listing.save!
+
+      expect(listing.name).to eq("Offer: grocery shopping, childcare")
+    end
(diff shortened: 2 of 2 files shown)


h-m-m:
@AngelaGuardia I like a nice clean solution like this

Since this change affects the UI, can you post "before" and "after" screenshots?
Also, do you mind adding a test for this? I'm happy to help with that if our testing setup isn't one you're familiar with

AngelaGuardia:
> @AngelaGuardia I like a nice clean solution like this
> 
> Since this change affects the UI, can you post "before" and "after" screenshots?
> Also, do you mind adding a test for this? I'm happy to help with that if our testing setup isn't one you're familiar with

@h-m-m I added the screenshots! I'm not sure how the map decides which listings to display (it usually just displays one). In this case, the one it displayed on main was different than the one on my branch but you can still see how the name is not shown anymore.

I would love to pair on adding tests! Both to this PR and the other I'm working on. I'm unfamiliar with the setup you're currently using.

h-m-m:
Great! The screenshots look good

I'm happy to go over our testing setup with you tomorrow during the project's usual Zoom call

exbinary (approved):
Sweet! 🍠 🙌🏾

PR #849, opened 2021-02-03
Admin param

exbinary:
### Why
Rolls the `?admin=false` logic into `ApplicationPolicy` via `Context` introduced in #844.

This will allow us to eventually drop the `@admin_status` controller variable and cleanly rely on policies instead.

Also serves as an example of how `Context` is useful outside of the controller layer.

### What and How

1. Adds a `admin_param` flag to Context, passed in from `ApplicationController`
1. Extract and use this in `ApplicationPolicy` to allow for `admin?` and `sys_admin?` privileges to be overridden by sending `?admin=false`
1. Adds new `can_admin?` methods to `ApplicationPolicy` and `Context`. This method tests whether the user is an admin or sysadmin and knows about the `?admin=false` url param override.

### Testing
- [x] `app/policies/application_policy_spec.rb`.
- [x] `app/models/context_spec.rb`

### Next Steps
?

?

### Pre-Merge Checklist

### DIFF
--- a/app/models/context.rb
+++ b/app/models/context.rb
@@ -1,4 +1,5 @@
 Context = Struct.new(
+  :hide_admin?,
   :system_settings,
   :user,
   keyword_init: true
--- a/app/models/context.rb
+++ b/app/models/context.rb
@@ -1,5 +1,5 @@
 Context = Struct.new(
-  :hide_admin?,
+  :admin_param,
   :system_settings,
   :user,
   keyword_init: true
--- a/app/models/context.rb
+++ b/app/models/context.rb
@@ -9,6 +9,10 @@ Context = Struct.new(
     self[:system_settings] ||= SystemSetting.current_settings
   end
 
+  def can_admin?
+    user && (user&.admin_role? || user&.sys_admin_role?) && admin_param != 'false'
+  end
+
   def to_h
     {
       system_settings: system_settings,
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -32,7 +32,7 @@ class ApplicationController < ActionController::Base
   def context
     @context ||= Context.new(
       user: current_user,
-      hide_admin?: params[:admin] == 'false',
+      admin_param: params[:admin],
     )
   end
   helper_method :context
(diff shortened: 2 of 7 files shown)


app/models/context.rb
```diff
@@ -7,4 +8,8 @@
   def system_settings
     self[:system_settings] ||= SystemSetting.current_settings
   end
+
+  def can_admin?
+    user && (user&.admin_role? || user&.sys_admin_role?) && admin_param != 'false'
+  end
```
h-m-m, line 14: I think I disagree with this. I'm fine with the PR going ahead with it, though, since I think it will be easy to edit later. 

Here's my thinking. I read your comment in the commit that says

> While this duplicates a bit of logic in ApplicationPolicy, it didn't
feel quite right to have to do `policy(model).change?` in a
controller to determine where to redirect to. The `.change?` part of
that call seems an arbitrary choice. In fact, while it seems related to
permissions, in this case what i'm asking is not whether the user has
permissions to do something but whether they are acting as an admin.

and compared that with how you use this method in https://github.com/rubyforgood/mutual-aid/blob/a6dac92e081ed3e5bd69b7e17400b4df49bf8b58/app/controllers/announcements_controller.rb#L50-L56

I can imagine a scenario where we want to do something like — "allow sys_admins to approve announcements but not regular admins"

If we use the policy(...).change? check in the controller sample I linked to in here, then we'd only need to set that logic in the policy, and the controller would automatically follow the new behavior. Whereas if we use `context.can_admin?` now we need to edit something — either `context.can_admin?` or something in the controller — to allow this new behavior.

Of course, this is a hypothetical change and I'm happy to worry about it later when we get there. So I'm fine with this being merged in even though what you call an arbitrary choice feels less arbitrary to me. I think this is probably influenced by my background in administering weird Windows Active Directory permission environments.

h-m-m (approved):
Thanks to @rodri0315 for help with reviewing this

app/models/context.rb
exbinary: Great points, i like all of that thinking. Pushing this logic into the policy, maybe with a custom query like `.review?` would make a lot of sense and be more future-proof as you describe. I think i was influenced by realizing several controllers need this kind of logic and so we'd need similar methods in several policies whereas `context.can_admin?` can be reused.

All that said, i think this is well worth keeping an eye on. As with most things, we'll learn as we use it more what works and what doesn't.

Appreciate the rigorous look! 💥.

PR #868, opened 2021-02-23
moved disable claim logic/fixed merge

svileshina:
### Why
Fixed the bad merge in 9de84a0d9f984120531227402f9e2124e1b58d00 in #846, moved the disabled claim logic into the new card

@leahriffell @AngelaGuardia - looks like this merge duplicated the cards. We went ahead and merged this but could you take a look to check that we brought everything over? 
Screen Shot 2021-02-23 at 10 01 38 AM


exbinary (approved):

leahriffell:
@svileshina @exbinary sorry about that! I checked and it does look like everything was brought over. Thank you for catching and removing that extra card in `show.html.erb`

PR #896, opened 2021-03-04
Drop credentials.yml.enc

exbinary:
### Why
Since our app supports many disparate deployments, most of our secrets aren't shared and so storing them in our repo doesn't really make sense.

The only secret currently in credentials is `secret_key_base` but even that is not used, AFAICT.
* In dev|test, rails [uses a randomly generated key](https://devdocs.io/rails~6.0/rails/application#method-i-secret_key_base) in `tmp/development_secret.txt`.
* In our heroku deployments, we use an ENV variable.

Moreover, credentials.yml.enc violates [12factor principles](https://12factor.net/config) and also often get out of sync and trip up the dev experience.

### What
- [x] Remove `config/credentials.yml.enc`
- [x] Add a comment and example in `.env` as a reminder to set this var on any new deployed env

### Next Steps
- [ ] Make sure this works on test

- [ ] Would this be a breaking change for any of our deployments?

### Security
I don't think this impacts our existing environmental security since we don't appear to have been using credentials.

### Pre-Merge Checklist


h-m-m:
Should this also be in the .gitignore?

maebeale (approved):

PR #898, opened 2021-03-05
add claim email text for p2p

maebeale:
Co-authored-by: connieh1 <connie2630a@gmail.com>
Co-authored-by: maebeale <maebeale@gmail.com>

### Why
Autoemails sent after someone Claims a Contribution needed some framing text

### What
We added a first pass of the Claim notification email template text

### Testing
N/A

### Next Steps
We should also consider sending an autoemail to the Claimer to confirm they did, in fact, Claim, and w note re that they should contact the mutual aid group if they don't hear back from their person within two days.

We prob will want to get out of this being a hard-coded option (or have an override).

### DIFF
--- a/app/mailers/peer_to_peer_match_mailer.rb
+++ b/app/mailers/peer_to_peer_match_mailer.rb
@@ -2,15 +2,15 @@
 
 class PeerToPeerMatchMailer < ApplicationMailer
   def peer_to_peer_email(contribution, peer_alias:, message:)
-    peer_email_address = contribution.person.email
+    @peer_email_address = contribution.person.email
     email_subject = if contribution.ask?
                       "#{peer_alias} is offering to meet your ask"
                     elsif contribution.offer? # TODO: check if community resource when added
                       "#{peer_alias} would like to accept your offer"
                     end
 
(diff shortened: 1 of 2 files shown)


app/mailers/peer_to_peer_match_mailer.rb
```diff
@@ -2,15 +2,15 @@
...
     email_subject = if contribution.ask?
                       "#{peer_alias} is offering to meet your ask"
                     elsif contribution.offer? # TODO: check if community resource when added
                       "#{peer_alias} would like to accept your offer"
                     end
 
-    mail(to: peer_email_address, subject: email_subject) do |format|
-      format.html { render locals: { message: message } }
+    mail(to: @peer_email_address, subject: email_subject) do |format|
+      format.html { render locals: { contribution: contribution, message: message } }
```
exbinary, line 13: I don't think i understand the distinction between using an instance var (`@peer_email_address`) vs. locals (`contribution`, `message`). The usage pattern in the template below looks pretty similar so wondering if we could consilidate on one or the other?

exbinary (approved):
Looks good! One minor question.

PR #903, opened 2021-03-12
Set default mailer url for production

mandelbro:
### Why

Resolves bug with confirmation email is incorrectly linking to port 38076

### What

- [x] Set default mailer url for production

### How

We determined that this problem exists both locally and on production, the prod server is likely using a non-standard port (38076) and that is being added to the confirmation url by the following line in `config/application.rb`:

```
    config.action_mailer.default_url_options = {
      host: ENV['SYSTEM_HOST_NAME'],
      port: (ENV['PORT'].presence || 80).to_i,
    }
```

Setting the host without the port should resolve this issue.

### Testing

There are currently no tests for this as it is a devise method, but this should be tested in a demo environment to ensure it works as expected.

### Next Steps

Test in the demo environment

None

### Accessibility

N/A

### Security
 
N/A

### Meta

N/A

### Pre-Merge Checklist

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -79,6 +79,9 @@ Rails.application.configure do
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
+  # Set default mailer url
+  config.action_mailer.default_url_options = { host: ENV['SYSTEM_HOST_NAME'] }
+
   # Inserts middleware to perform automatic connection switching.
   # The `database_selector` hash is used to pass options to the DatabaseSelector
   # middleware. The `delay` is used to determine how long to wait after a write


mandelbro:
For reference, there are several ways to set this default url, as outlined here: https://github.com/heartcombo/devise/wiki/How-To:-Set-:host-and-:port-for-all-devise-mailer-urls, but we decided setting in the production config was the lightest weight way to implement the fix.

mandelbro:
Here's how we set it on our app

```


    default_url_options = {
      host: ENV.fetch('HOSTNAME'),
      locale: I18n.locale,
      protocol: ENV.fetch("URI_SCHEME")
    }

    config.action_mailer.default_url_options = default_url_options
```

Where the env variable is set to `define HOSTNAME="localhost:3000"` locally, and set to our production url on prod.

maebeale (approved):
YES!

PR #897, opened 2021-03-04
Issue #782 : Ruby version upgraded to 2.7.2.

sharvy:
### Why
To include security fix in Webrick and many [other changes](https://github.com/ruby/ruby/compare/v2_7_1...v2_7_2).
Resolves [#782](https://github.com/rubyforgood/mutual-aid/issues/782)

### How
- [x] Update Gemfile and ruby-version files
- [x] Upgraded Node to 14+ and Webpacker to 5+. 
- [x] Update Development Dockerfile & Production Dockerfile
- [x] Change CircleCI Config
- [x] Change Ansible Config 
- [x] Build locally, run tests `bin/dev/test`, make sure all Green
- [x] Make sure CircleCI is working

### Pre-Merge Checklist
- [x] Documentation and comments have been added to the codebase where required
- [x] Outstanding questions and concerns have been resolved

PR #905, opened 2021-03-17
Fix local webpack compilation

exbinary:
### Why
It appears that #897 broke webpack compilation locally (dev, test), at least on my machine, failing with:
`Error: Node Sass does not yet support your current environment: OS X 64-bit with Unsupported runtime (83)`

Some research led me to https://github.com/rails/webpacker/pull/2716 which has been merged into webpacker 6 but not backported into our current version.

### How
This changeset manually overrides the sass-loader implementation from `node-sass` to `sass`, as described [in this article](https://mentalized.net/journal/2019/10/19/use-sass-modules-in-rails/).

### Next Steps
Left a FIXME comment in `config/webpack/environment.js` to remove this workaround when we upgrade to webpacker 6.

?

### Pre-Merge Checklist

### DIFF
--- a/config/webpack/environment.js
+++ b/config/webpack/environment.js
@@ -2,6 +2,17 @@ const path = require('path')
 const { environment } = require('@rails/webpacker')
 const { VueLoaderPlugin } = require('vue-loader')
 
+// Override sass-loader implementation from deprecated node-sass to dart sass
+// https://mentalized.net/journal/2019/10/19/use-sass-modules-in-rails/
+// FIXME: remove this workaround when webpacker is upgraded to 6+
+environment
+  .loaders
+  .get('sass')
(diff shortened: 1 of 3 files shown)


sharvy:
@exbinary This looks great! +1 for `dart-sass`. 

I think you should create the issue *Upgrade to Webpacker 6* and mention about this patch, so that whoever takes that issue, instantly knows about it.

exbinary:
> I think you should create the issue _Upgrade to Webpacker 6_ and mention about this patch, so that whoever takes that issue, instantly knows about it.

Good thinking! Tracked in #906

exbinary:
Merging this w/o full approval because AFAICT, `main` is currently broken for local development.

PR #923, opened 2021-04-03
Improve locations on community resources

exbinary:
### Why
Primarily makes location fields optional on the `CommunityResource` form and ensures a 1-1 relationship betweeen CommunityResources and their Locations. Also introduces some new tools and patterns for working with Form objects.

### What

- [x] Drop location lookup/reuse in `CommunityResourcesController#create`. The current implementation attempts to find a matching `Location` record and reuse it. In discussion with @maebeale and @h-m-m, decided to instead treat our `locations` table as merely an extraction of columns that would otherwise be duplicated in multiple host tables. The `Location` model and table give us a way to extract logic and columns, but do _not_ represent anything more semantic meaning. In particular, location records are not intended to be unique or reused; each one is completely owned and tied 1 to 1 with a single host record.
- [x] Make location fields optional
- [x] Introduce a `CommunityResourceForm` to simplify the controller and ease unit testing. The form is reused in the create and updated actions.
- [x] Introduce a `BaseForm.filter_keys` method to eliminate duplication between form objects and pundit policies. Both need a list of expected attributes: the form defines these as filters and the policy in `permitted_attributes`. `filter_keys` is designed to be called by the policy and generates a strong_params compatible list of all filters defined on the form.

### Testing
- [x] Tried to cover important scenarios in `community_resource_form_spec.rb`.
- [x] Moved some existing/obsolete specs out of `requests/community_resource.rb`.
- [x] Added specs to `base_form_spec.rb` for new `.filter_keys` functionality.

### Next Steps
- [x] Need to clean up one todo: `CommunityResourceForm` handles its nested `organization` and `location` differently. One uses `accepts_nested_attributes` while the other uses a nested form object. Would be good to be consistent with one or the other -- marked as a #FIXME.
- [x] Needs some more manual testing.

?

### Security
- [x] Needed to ensure the changes here didn't allow non-admins to toggle whether a CR is approved.

### Pre-Merge Checklist

### DIFF
--- a/app/controllers/community_resources_controller.rb
+++ b/app/controllers/community_resources_controller.rb
@@ -17,20 +17,8 @@ class CommunityResourcesController < ApplicationController
   end
 
   def create
-    location_params = params['community_resource']['location']
-
-    location = Location.where(
-      street_address: location_params['street_address'],
-      city: location_params['city'],
-      state: location_params['state'],
-      zip: location_params['zip'],
-      location_type_id: location_params['location_type_id']
-    ).first_or_create
-
-    community_resource.location = location
-    community_resource.update permitted_attributes(community_resource)
-
-    if community_resource.save
+    @community_resource = authorize CommunityResourceForm.build params[:community_resource]
+    if @community_resource.save
       redirect_after_create
     else
       render :new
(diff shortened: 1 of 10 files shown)


maebeale (approved):
YESSSS!!!!!

exbinary:
Apologies @leahriffell -- i forgot to tag you on this!
Figured you'd want to see how the effort you started turned out. Thanks again! 🙌🏾

Issue #240, opened 2020-05-14
[DISCUSSION] Organization.current_organization


maebeale:
- This shouldn't be renamed to `is_instance_owner`, bc it'll get also connected to an admin user's `organization`, so, when 4 groups are sharing the system, it'll default to the correct group.

- Imo, on the Vue side we'll want to have the org as a URL param rather than only being able to access org name via `Organization` model

- I liked the name `current_organization` bc of its nod to `current_user`

https://github.com/maebeale/mutual-aid/blob/8ec5e011c71d7adef629ea11b15c49c0ed35896d/app/models/organization.rb#L10

PR #926, opened 2021-04-06
Ensure only one `current_organization` and rename to `instance_owner`

exbinary:
### Why
Discovered that every organization currently created is defaulted to `is_instance_owner: true`.
IIUC, only one org should be the instance owner.

Also renames `current_organization` because that name connotes something transient, eg `current_user`.

### What
- [x] Migration to change `organizations.is_instance_owner` default to false
- [x] Model validation to ensure no two organizations can be instance owners
- [x] Rename `Organization.current_user` -> `.instance_owner`
- [x] Change `Organization.instance_owner` from a scope to a class method since it doesn't return a relation

### Testing
- [x] Added a model spec
- [x] Manually tested org form

### Next steps
- [x] Disable`is_instance_owner` attribute on form and add a note. (instance owner has to be seeded)
- [x] Add an issue to the roadmap to support multiple hosts using the same instance. #930 

It's quite possible i'm completely misunderstanding the intention of `current_organization`!
@maebeale, how does this align with your initial/current thinking?

### Accessibility
This does make it harder for admins to change which org is the instance owner. If that becomes a needed feature, we can add a 'transfer ownership' feature.

### Pre-Merge Checklist


exbinary:
Whelp, i just noticed #240, where @maebeale explicitly says this should _not_ be renamed :)
Let's discuss!

maebeale:
Hey! I'm ok with this!

PR #927, opened 2021-04-08
Free specs from seeding

exbinary:
### Why
Our ruby specs currently depend on the test database being seeded, which is non-standard and surprising, particularly for new contributors.

### What, How
- [x] Consolidate `logo_url` logic into a new helper method and make it defensive in case the DB doesn't have an `instance_owner`.
    - [x] Moves existing logic out of `shared/_organization_logo` and `NavBar.vue`.
- [x] Extract `GenerateLandingPageJson` interactor out of `PublicPagesController.landing_page` to ease testing.
- [x] Create `LocationType` in a couple of specs instead of relying on one having been seeded.
- [x] Thread `SystemSetting` into `SubmissionMailer` so it can be injected easily in specs.
- [x] Thread `Organization` into `EmailNewSubmission` and `EmailPeer` interactors, and the corresponding mailers (`SubmissionMailer`, `PeerToPeerMatchMailer`)
- [x] Drop seeding from CircleCI rspec task.

### Testing
Relies on existing specs and some manual testing of nav bar, login and signup pages.

### Next Steps
- [ ] Probably worth exploring whether `Organization.instance_owner` should live in `Context`, similar to `SystemSetting.current_settings`.
- [x] Remove mention of needing to seed in docs.

Note this is built on top of #926 which renamed `Organization.current_organization` to `.instance_owner`. However the changes here aren't really dependent on that particular change.

### Pre-Merge Checklist
- [x] Entry to CHANGELOG.md not necessary
- [x] Any next steps have been turned into Issues or Discussions as appropriate

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -73,4 +73,9 @@ module ApplicationHelper
       urgency_level_text
     end
   end
+
+  def site_logo_url
+    # There should always be a current org, but being defensive here helps simplify tests
+    Organization.instance_owner&.logo_url.presence || asset_pack_path('media/images/logo.png')
+  end
 end
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ commands:
           command: |
             set -euxo pipefail
             bundle config path .bundle
-            bundle exec rails db:setup
+            bundle exec rails db:create
             bundle exec rspec --format progress --format RspecJunitFormatter -o ~/rspec/rspec.xml
   yarn_tests:
     description: "Run the Mocha test suite"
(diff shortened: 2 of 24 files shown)


maebeale:
Love this: `Probably worth exploring whether Organization.instance_owner should live in Context, similar to SystemSetting.current_settings.`

maebeale (approved):

app/helpers/application_helper.rb
```diff
@@ -73,4 +73,9 @@ def urgency_level_display(urgency_level_text)
       urgency_level_text
     end
   end
+
+  def site_logo_url
+    # There should always be a current org, but being defensive here helps simplify tests
+    Organization.instance_owner&.logo_url.presence || asset_pack_path('media/images/logo.png')
```
h-m-m, line 79: Thank you! I've tripped up on this many times, and this feels to me like a much better home for this conditional than anywhere else it has lived before

Issue #885, opened 2021-03-03
Update Announcement Show to be more palatable to the public


maebeale:
* Remove the label/data format
* Make title be the header (instead of Show Announcement)
* Display Description (or "(no description)")
* Remove date range and instead add note re when it will expire

hairedfox:
@maebeale I'd love to take this issue. Thank you

hairedfox:
@maebeale 
> Remove the label/data format
Does this mean that you want to remove all the labels?

> Make title be the header (instead of Show Announcement)
Is it `annoucement.name`? 

> Remove date range and instead add note re when it will expire
Could you explain more about this line? Thank you

exbinary:
I believe this was addressed in #928 .

PR #928, opened 2021-04-08
make Announcement show page like a card

hairedfox:
### Issue
#885 

### Why
- Make the Announcement show page like a card, make it fancy

### What
- Add some Bulma style for Announcement show page

### Screenshots
image

--Without edit permission:
image

### DIFF
--- a/app/presenters/announcement_presenter.rb
+++ b/app/presenters/announcement_presenter.rb
@@ -1,0 +1,8 @@
+class AnnouncementPresenter < SimpleDelegator
+  def expires_on
+    return "This announcement does not expire." unless publish_until
+
+    expire_text = self.publish_until > Time.current ? "Expires" : "Expired"
+    "#{expire_text} on #{I18n.l(publish_until, format: :long_month)}"
+  end
+end
--- a/app/views/announcements/show.html.erb
+++ b/app/views/announcements/show.html.erb
@@ -1,31 +1,26 @@
-<%= render "layouts/view_header", resource: announcement %>
-
+<% announcement_presenter = AnnouncementPresenter.new(announcement) %>
 <p id="notice"><%= notice %></p>
+<div class="card">
+  <header class="card-header">
+    <p class="card-header-title title is-3">
+      <%= announcement_presenter.name %>
+      <% if policy(announcement).change? %>
+        <%= link_to 'Edit', edit_announcement_path(announcement), class: "button is-primary ml-1" %>
+      <% end %>
+    </p>
+  </header>
+  <div class="card-content">
+    <div class="content">
+      <p>
+        <%= announcement_presenter.description %>
+      </p>
 
-<div class="section-detail">
-  <p>
-    <strong>Name:</strong>
-    <%= announcement.name %>
-  </p>
-
-  <p>
-    <strong>Description:</strong>
-    <%= announcement.description %>
-  </p>
-
-  <p>
-    <strong>Display date start:</strong>
-    <%= announcement.publish_from %>
-  </p>
-
-  <p>
-    <strong>Display date end:</strong>
-    <%= announcement.publish_until %>
-  </p>
-
-  <% if policy(announcement).change? %>
-    <%= link_to 'Edit', edit_announcement_path(announcement) %> |
-  <% end %>
-  <%= link_to 'Back', announcements_path %>
+      <p>
+        <small><em><%= announcement_presenter.expires_on %></em></small>
+      </p>
+    </div>
+  </div>
+  <div class="card-footer">
+    <%= link_to 'Back', announcements_path, class: "button is-info ml-1 mt-1 mb-1" %>
+  </div>
 </div>
-
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -53,3 +53,6 @@ en:
       zero: Match Feedback
   pundit:
     default: 'You cannot perform this action.'
+  date:
+    formats:
+      long_month: "%B %d, %Y"


hairedfox:
@maebeale when you have time, please take a look at this :D Thank you

exbinary (commented):
Thanks for working on this @hairedfox, it's looking great! 🔥

One minor comment inline and i have some thoughts about `draper`.

In general, i love decoration as a pattern to uphold the Single Responsibility Principle, keeping extraneous logic out of models and views. I also tend, as you've done here, to stay away from view helpers.

So i think we're in agreement that our codebase needs a clean place for view logic and that decoration is a helpful tool to help with this.

That said, i question whether `draper` is the best solution. There's nothing technically wrong with it, but it comes with some conventions that don't hold up very well in the long run. In particular, the `.decorate` helper encourages a 1 to 1 correspondence between models and decorators (eg `announcement.decorate` returns a `AnnouncementDecorator`). This is fine at first, but i've found in past projects that decorators start to bloat as they accrue methods needed across several different views. Over time, they become a grab-bag of methods called from various places, much like, eg, an `announcements_helper` module would.

I think part of what has happened in the rails world is that `draper` became a common choice for implementing the Presenter pattern, because it allows for easy decoration. But the subtle convention built into draper -- 1 model to 1 decorator -- actually ends up leading us _away_ from true Presenters.

Here's a great [article](https://nithinbekal.com/posts/rails-presenters/) describing an implementation of Presenters that addresses the same concerns, with some subtle differences:
* Uses the standard ruby `SimpleDelegator` for delegation instead of needing a gem like `draper`.
* Encourages us to name the decorators accurately, ie `AnnouncementPresenter` instead of `AnnouncementDecorator`. This subtle difference helps remind us of the purpose of this layer of objects.
* The initializer in `BasePresenter` explicitly accepts both the model instance and a `ViewContext` instance. With `draper`, this view context is made available 'magically', making it harder to mock out during testing.
* The `present` helper introduced near the end uses the same convention as `draper` does, looking for a 1-1 match between model and presenter. But we could choose a different convention here, such as one based on the controller action. Or even forego the automatic lookup altogether and require the presenter to be explicitly given. This flexibility helps us avoid the 1-1 trap described above.

Ok, i'm not sure how well i've articulated my thoughts on this :).
Please let me know your thoughts and i'm happy to keep discussing.

Thanks again for your work on this! 🙏🏾 

/cc @h-m-m because i think we've discussed presenters briefly in the past.

app/decorators/announcement_decorator.rb
```diff
@@ -0,0 +1,8 @@
+class AnnouncementDecorator < ApplicationDecorator
+  def expire_until
```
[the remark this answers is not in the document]
hairedfox: @exbinary ohh, sorry for my bad English 😓 . Yeah, `expires_on` makes more sense

hairedfox:
@exbinary Thank you for informing. I think Presenter is a good solution also, and it doesn't involve any gems or hierarchy that we have to follow.

I'll create Presenter for each controller, I don't know if it matches your idea or not but here it is 

```ruby
# app/presenters/announcement_presenter.rb
class AnnouncementPresenter
  def initialize
    # ...
  end

  def expires_on(announcement)
    # logic goes here
  end
end

# app/controllers/announcements_controller.rb
class AnnouncementsController < ApplicationController
  def some_method
    @presenter = AnnouncementPresenter.new
  end
end
```

```erb
<%# app/views/announcements/some_method.html.erb %>
<%= @presenter.expires_on(announcement) %>
```

Just let me know how you think about this. Thank you for opening this discussion :D I'd love to discuss more about this.

exbinary:
Thanks for the quick response! Yes, that's in the direction i'd go as well.

Expanding on your code, i'd experiment with something like this (based on the article mentioned above):
```ruby
# app/presenters/base_presenter.rb
class BasePresenter < SimpleDelegator
  def initialize(model, view_context)
    @view_context = view_context
    super(model)
  end

  # allows access to helper methods, eg h.time_ago_in_words
  def h
    @view_context
  end
end

# app/presenters/announcement_presenter.rb
class AnnouncementPresenter < BasePresenter
  def expires_on
    # ...
  end
end

# app/helpers/application_helper.rb
  # ...
  def present(model, presenter_class)
    presenter = presenter_class.new(model, self)
    block_given? ? yield(presenter) : presenter
  end
  # ...
```

Which then allows this pattern in templates:
```erb
# app/views/announcements/show.html.erb
<% present(announcement, AnnouncementPresenter) do |announcement| %>
    <%= announcement.name %>
```

exbinary:
> I'll create Presenter for each controller, I don't know if it matches your idea or not but here it is

Oh, and just to be clear, i don't think we need to create presenters for every controller or model. We can introduce them as needed.

hairedfox:
@exbinary @maebeale Could you guys take a look at this again? Thank you :bow:

app/decorators/base_decorator.rb
```diff
@@ -0,0 +1,17 @@
...
+    yield(self) if block_given?
+  end
+
+  def method_missing(method_name, *args)
+    obj.public_send(method_name, *args)
+  end
+
+  def respond_to_missing(method_name, include_private = false)
+    super
+  end
```
exbinary, line 16: We could also use `SimpleDecorator` which implements this for us. I believe it also tries to mitigate the performance penalty incurred by `method_missing` (eg. by defining methods lazily so that a second invocation meets a real method).

app/helpers/application_helper.rb
```diff
@@ -78,4 +78,10 @@ def site_logo_url
     # There should always be a current org, but being defensive here helps simplify tests
     Organization.instance_owner&.logo_url.presence || asset_pack_path('media/images/logo.png')
   end
+
+  def present(obj, decorator, &block)
+    decorator.public_send(:new, obj, &block)
```
exbinary, line 83: I think we could call `decorator.new` directly here instead of needing `public_send`?

app/helpers/application_helper.rb
```diff
@@ -78,4 +78,10 @@ def site_logo_url
...
+
+    nil
+  end
```
exbinary, line 86: I'd like to also return the presenter here so that views can also choose to assign the result of `present` to a variable and use it later, eg:
```erb
<% presenter = present(@announcement, AnnouncementPresenter) %>
...
<%= presenter.name %>
```

exbinary (commented):
Thanks for the updates @hairedfox ! Looking better and better 🚀 .
Just a few more comments...!

hairedfox:
@exbinary Thank you for your review. I've been a little busy for a day or two so I'll change this tomorrow. About `SimpleDecorator`, do we actually need that?

exbinary:
> @exbinary Thank you for your review. I've been a little busy for a day or two so I'll change this tomorrow.

No worries! Thanks again for working on this and being open to discussion!

> About `SimpleDecorator`, do we actually need that?

Oops, sorry i mistyped that -- i meant `SimpleDelegator` ([docs](https://devdocs.io/ruby~2.7/simpledelegator)).
I agree a gem would seem like overkill, but since this is already part of ruby, i'd prefer to use it instead of writing our own `method_missing` and related hooks. There are edge-cases to worry about when implementing that kind of thing.

hairedfox:
@exbinary I've just pushed the fix. Please review this when you have time. Thank you in advance. :bow:

exbinary (approved):
This looks great! 🙏🏾

PR #931, opened 2021-04-09
Force Warden.test_reset! after request specs

exbinary:
Happened to notice this in `doc/testing.md`:
> BE SURE to include the line `after { Warden.test_reset! }` after the `before do` block with the `login_as` method within it. This prevents unexpected state data of the user from hanging around and causing errors.

I'm pretty sure none of us are doing that 😅 !

This PR adds an `after` hook to `rails_helper.rb` that automatically resets `warden` for request specs.


maebeale (approved):
Lol.!!!!

PR #932, opened 2021-04-10
Provide access to `Organization.instance_owner` from `Context`

exbinary:
### What
Provides access to `Organization.instance_owner` from the `Context` object introduced in #844.

### Why
`instance_owner`, like `SystemSetting.current_settings` is a request-scoped global that is better represented as part of `context` than by a class accessor. `Context` is also easier to inject and therefore mock in specs.

Note that i've named the accessor `host_organization` instead of sticking with `instance_owner`.

While `instance_owner` makes intuitive sense when hung on `Organization`, i felt that `context.instance_owner` isn't as clear. If the reader doesn't already know the term, the method name is a bit cryptic and doesn't give any indication of the type of object returned. Since this method is accessed in several places far from `Organization`, `context.host_organization` seemed more transparent.

The downside is that we now have two terms in the codebase that refer to the same thing, one within the `Organization` class and table, and another everywhere else. We could take this thinking a bit further and rename `Organization.instance_owner` to `.host_organization`, leaving only the table column name as `is_instance_owner`.

Definitely open to thoughts about this (it's easy enough to rename either way).

### Pre-Merge Checklist

- [x] Entry added to CHANGELOG.md not necessary


app/models/context.rb
```diff
@@ -15,6 +20,8 @@ def can_admin?
 
   def to_h
     {
```
maebeale, line 22: Shouldn't this have already been in here? Or, we weren't calling it before?

spec/models/context_spec.rb
```diff
@@ -38,11 +38,11 @@
...
 
   describe 'to_h' do
-    let(:user) { double 'User' }
-    let(:system_settings) { double 'SystemSettings' }
+    let(:user) { instance_double User, 'admin_role?' => true }
+    let(:system_settings) { instance_double SystemSetting }
     let(:context) { Context.new user: user, system_settings: system_settings }
 
-    def method_with_keyword_args positional_arg, non_context_keyword_arg:, user:, system_settings:
+    def method_with_keyword_args positional_arg, non_context_keyword_arg:, user:, system_settings:, **_rest
```
maebeale, line 45: Why is this needed?

spec/models/context_spec.rb
```diff
@@ -58,7 +58,7 @@ def method_with_keyword_args positional_arg, non_context_keyword_arg:, user:, sy
     subject { context.can_admin? }
 
     context 'sysadmin' do
-      let(:user) { instance_double 'User', admin_role?: false, sys_admin_role?: true }
+      let(:user) { instance_double User, admin_role?: false, sys_admin_role?: true }
```
maebeale, line 61: Do we want both `can_admin?` and `admin_role?`

maebeale:
I'd be fine wth changing the db name too! :)

spec/models/context_spec.rb
maebeale: And, nice fix! Do we need an extra test to make sure this works (since we didn't know this was broken before)?

spec/models/context_spec.rb
exbinary: Great question! I didn't mention anything about this in the commit or PR -- it's a bit irrelevant and i forgot about it :).

This spec itself is a bit confusing IMO, and maybe not even necessary. I was experimenting with how to easily destructure `context` into method call args, for the cases where the method needs multiple context properties.

The destructuring works nicely with `.to_h`, but in order for it to be really resilient, the method has to accept a 'glob' arg to cover all the other context attributes it doesn't care about. In this case, adding the new `host_organization` attribute to `Context.to_h` broke this spec -- because `method_with_keyword_args` wasn't expecting it.

maebeale (approved):

exbinary:
Renamed the `is_instance_owner` column to `is_host`. Please give it one more look @maebeale .

maebeale:
(the PR text should change too, right?)

app/controllers/asks_controller.rb
```diff
@@ -16,9 +16,9 @@ def create
     if submission.save
       EmailNewSubmission.run!(
         submission: submission,
-        user: current_user,
-        system_setting: context.system_settings
-        organization: Organization.instance_owner
+        user: context.user,
+        system_setting: context.system_settings,
```
maebeale, line 20: 👍

maebeale (approved):
looks great! love having this all match now!!!

PR #935, opened 2021-04-14
Renew RuboCop effort based on StandardRB

exbinary:
### Why
Back in October, @smoline put in some great work introducing `rubocop` into our codebase.
Sadly, we had to prioritize feature development and weren't able to complete and merge all that work.
This is the first of several PRs intended to revive that effort.

Since then, [StandardRB](https://github.com/testdouble/standard) has hit v1.0 and i think there's a lot to gain from basing our rules on theirs. Primarily, they've done a lot of work to weigh different alternative options and tried to standardize, which means we have less to think about!

That said, StandardRB is an inherently 'opinionated' tool and (not surprisingly 😏), i would like a little more flexibility than it allows.

As such, i'm proposing that we **base** our rules on StandardRB but retain the ability to customize on top of them.

### What
In this PR:
- [x] Replace our fledgling `.rubocop.yml` config with one that inherits from StandardRB, via the `standard` gem.
- [x] Generate a `.rubocop_todo.yml` which encapsulates and excludes all existing violations. This allows us to enable linting of new code without having to first fix the whole codebase. It also sets up follow-on PRs to fix violations; see Next Steps below.
- [x] Cherry pick @smoline's 62ce3a0, which adds rubocop to our CircleCI build 💥🙏🏾

See Next Steps for a list of follow-up PRs that customize the ruleset and incrementally address existing violations.

### How
The integration with StandardRB is based on this [evilmartians.com article](https://evilmartians.com/chronicles/rubocoping-with-legacy-bring-your-ruby-code-up-to-standard#one-standard-to-rule-them-all).

The idea is to continue using `rubocop` executables instead of `standard` but simply to base our rubocop config on theirs.

### Testing
Adds StandardRB-based rubocop validation to our CI builds.

### Next Steps
- [x] The following PRs build on this one to address existing violations, incrementally paring down `.rubocop_todo.yml`. Some of them also customize or override Standard rules. Most build on the previous one and can be merged independently in the order listed below.
    - [x] #936 Rules related to spaces
    - [x] #937 Indentation consistency
    - [x] #938 Alignment and Indentation
    - [x] #939 Trailing commas
    - [x] #940 Single line methods
    - [x] #941 String literals
    - [x] #942 Empty lines and trailing whitespace
    - [x] #943 Remaining style rules
    - [x] #944 Semantic blocks
    - [x] #947 Unsafe YAML.load
    - [x] #948 Linting
    - [x] #949 Metrics
[...]

PR #938, opened 2021-04-15
Rubocop rules related to indentation and alignment

exbinary:
### Why
Building on #935, enables rules related to indentation and alignment, fixing corresponding violations.

### What
See the commit list for rules tackled in this PR. The commit comments include links to the corresponding rubocop docs.

Some rules worth highlighting:
* Disables one StandardRB rule: [`Layout/HashAlignment`](https://docs.rubocop.org/rubocop/cops_layout.html#layouthashalignment). See commit comments for my rationale and discussion c00c9bb.
* StandardRB configures [`ArgumentAlignment`](https://docs.rubocop.org/rubocop/cops_layout.html#layoutargumentalignment) in the `with_fixed_indentation` style, which means any multi-line args will be indented relative to the first line, _not_ aligned with the first argument. I found this worked well in our codebase but curious if anyone has differring opinions. See 1cae87d for examples.
* Similarly ^^ for [`ArrayAlignment`](https://docs.rubocop.org/rubocop/cops_layout.html#layoutarrayalignment).
* I find [`FirstArgumentIndentation`](https://docs.rubocop.org/rubocop/cops_layout.html#layoutfirstargumentindentation) in the `consistent` style recommended by StandardRB to be a bit stifling, but not enough to override.
* In the StandardRB style for [`MultilineMethodCallIndentation`](https://docs.rubocop.org/rubocop/cops_layout.html#layoutmultilinemethodcallindentation), multi-line method chains are indented two spaces relative to line of the first receiver. This is different from a common style in our codebase where lines are aligned on the 'dots'. Both styles have tradeoffs so i've stuck with StandardRB on this. See ebfa676 for examples.
* StandardRB recommends the `symmetrical` style for [MultilineMethodCallBraceLayout](https://docs.rubocop.org/rubocop/cops_layout.html#layoutmultilinemethodcallbracelayout). There are a few places where it's nice to be able to keep one or two args on the same line as the method call but put the rest on separate lines. Good examples are `FactoryBot.create` calls and AR `has_many` callls. In these cases, i've been able to get around this rule by using explicit braces around the trailing hash arg. See 607ec35 for examples.

Note: here's the [StandardRB rubocop config](https://github.com/testdouble/standard/blob/master/config/base.yml).

### How
Please review each commit separately and see commit comments for any related explanation/detail.

### Testing

### Next Steps
?

Open to thoughts about any of the rules enabled here!

### Pre-Merge Checklist


app/controllers/matches_controller.rb
```diff
@@ -61,11 +59,12 @@ def update
     update_connections = params[:commit]&.downcase&.include?('edit who this match connects')
     if @match.update(match_params)
       if save_and_continue || update_connections
-        redirect_to edit_match_path(@match, edit_connection_mode: update_connections ? true : false),
-                    notice: 'Match was successfully updated.'
+        redirect_to(
+          edit_match_path(@match, edit_connection_mode: update_connections ? true : false),
```
h-m-m, line 63: @exbinary — I didn't think any of the rules you mentioned in the discription would add the `(` after `redirect_to` here. Do you know which one does this?

exbinary: Ah yes, good question! This is something i tweaked manually. The rule at work here is `Layout/ArgumentAlignment`, which forces any multi-line arguments to be indented one level after the method call, not aligned under the first argument as in the original code. But with that automated change, i find the code to be less scannable:
```ruby
# original form -- disallowed by Layout/ArgumentAlignment
redirect_to edit_match_path(@match, edit_connection_mode: update_connections ? true : false),
            notice: 'Match was successfully updated.'

# autocorrected form -- i find this to be imbalanced and harder to scan
redirect_to edit_match_path(@match, edit_connection_mode: update_connections ? true : false),
  notice: 'Match was successfully updated.'

# balanced form -- i prefer this form where the first argument is on a newline, but it requires parentheses
redirect_to(
  edit_match_path(@match, edit_connection_mode: update_connections ? true : false),
  notice: 'Match was successfully updated.'
)
```
Does that make sense? Curious your thoughts on this.

h-m-m: In discussing this further with @exbinary, I realized my objection was that I keep misreading this as a three-argument call to `edit_match_path` that is indented weirdly. I have no objection to the guideline that's used here, and my confusion is more due to the ternary operator or the fact that `redirect_to` is multi-line when its nested `edit_match_path` isn't

exbinary:
Reviewed in person with @h-m-m and @svileshina .

PR #943, opened 2021-04-16
Remaining rubocop style rules

exbinary:
## Why
Building on #935, fixes violations of remaining Style rules.

### What
See commits and commit comments for list of full rules tackled.

Includes disabling of [`Style/SafeNavigation`](https://docs.rubocop.org/rubocop/cops_style.html#stylesafenavigation).
Standard enables this rule but i'm recommending we don't.

When a chain of calls needs to be guarded against a `nil`, it can often be meaningful to explicitly highlight which part of the chain might be nil. While the first `&.` makes this evident, i find the explicit form to be easier to spot.
```ruby
# what Standard recommends:
user&.person&.name # i find it harder to quickly spot what's going on here
# vs
user.person&.name if user.person # this is more explicit
```
The explicit form is also a bit more cumbersome, which hopefully creates some motivation to consider refactoring away from the need to guard against this.

In another example, code such as the following would lose significant meaning if corrected according to this rule:
```ruby
user.person.destroy! if user.person
# vs
user.person&.destroy!
```
Note that i'm not suggesting we _always_ use the explicit form. More that by disabling this rule, we retain the ability to choose which form is more expressive in a given situation

### Testing

### Next Steps
?

Really curious the team's thoughts on this.

### Pre-Merge Checklist


h-m-m:
> Really curious the team's thoughts on this.

I really haven't thought about this much. Because of that, I'm fine if you're leaning toward flexibility for now and changing our mind later

svileshina (approved):

PR #945, opened 2021-04-17
More filtering cleanup

h-m-m:
### Why

I wanted to try adding a filter now that we've refactored how filters work in #546

I also wanted to try @exbinary's suggestion about a refactor to how parameters are passed around. I feel like this refactor is not yet complete — for example we might want to look into some of the duplication that now exists — but I felt this is a good point to stop and ask for feedback before going further

### What

Screen Shot 2021-04-17 at 11 36 27 AM

### Pre-Merge Checklist

### DIFF
--- a/app/models/browse_filter.rb
+++ b/app/models/browse_filter.rb
@@ -3,16 +3,15 @@
 # BrowseFilter lets the browse page filter contributions by parameters, delegating much of the actual filtering
 # to the classes listed in BrowseFilter::FILTER_CLASSES
 class BrowseFilter
-  FILTERS = {
-    'Category' => CategoryFilter,
-    'ServiceArea' => ServiceAreaFilter,
-    'ContactMethod' => ContactMethodFilter,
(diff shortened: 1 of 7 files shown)


app/models/browse_filter.rb
```diff
@@ -3,16 +3,15 @@
...
-    hash[key] = {}
+  FILTER_CLASSES = [
+    CategoryFilter,
+    ServiceAreaFilter,
+    ContactMethodFilter,
+    ContributionTypeFilter,
+    UrgencyFilter
+  ].freeze
+  ALLOWED_PARAMS = FILTER_CLASSES.each_with_object({}) do |filter, hash|
+    hash.merge! filter::ALLOWED_PARAMS
```
exbinary, line 14: As i look at this, i'm wondering if we even need `ALLOWED_PARAMS` anymore? Since each filter will extract only the params it needs, maybe `ContributionsController` doesn't need to invoke strong_params at all? Perhaps that addresses some of the duplication you mentioned?

PR #946, opened 2021-04-20
Fix bug in Claiming introduced in acd210de08

exbinary:
### Why
Fixes a [bug](https://mutual-aid-app.slack.com/archives/C01LSKPJZNW/p1618666427000700) in peer-to-peer claims.

### What
acd210de08 introduced a bug where `Organization` was added as a required argument to `EmailPeer`, but not
supplied by `ClaimContribution` and `ClaimsController`.

### Testing
This bug went uncaught because `claim_contribution_spec` was mocking EmailPeer. Changing the spec to let `ClaimContribution` call composite interactions as it normally would surfaces any such integration problems.

### Next Steps
?

?

### Pre-Merge Checklist

### DIFF
--- a/spec/interactions/claim_contribution_spec.rb
+++ b/spec/interactions/claim_contribution_spec.rb
@@ -1,32 +1,34 @@
 require 'rails_helper'
 
 RSpec.describe ClaimContribution do
-  before do
-    allow(MatchContribution).to receive(:run!)
-    allow(EmailPeer).to receive(:run!)
-  end
-
   let(:contribution) { create(:listing) }
   let(:user) { create(:user, :with_person, email: 'user_email@example.com') }
 
-  before do
+  let(:claim_contribution) do
     ClaimContribution.run!(
       contribution: contribution.id,
       current_user: user,
+      organization: build(:organization),
       peer_alias: 'alias',
       message: 'message',
     )
   end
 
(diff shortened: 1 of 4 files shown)


spec/interactions/claim_contribution_spec.rb
```diff
@@ -1,32 +1,34 @@
...
       contribution: contribution.id,
       current_user: user,
+      organization: build(:organization),
       peer_alias: 'alias',
       message: 'message',
     )
   end
 
-  it "matches contribution and sends an email to peer", :aggregate_failures do
-    expect(MatchContribution).to have_received(:run!)
```
maebeale, line 22: Do we use this elsewhere?

exbinary: Do you mean `MatchContribution`?

maebeale: I meant the first, but was wondering re both while reviewing from my phone. :)

exbinary: Thanks! If you're asking about testing, this example is replaced by the two below and changed from being mocked to checking the side-effects of `MatchContribution` and `EmailPeer`.

As for production code, both are only used in `ClaimContribution`.

svileshina (approved):

PR #947, opened 2021-04-20
Introduce ToBoolean refinement to replace YAML.load

exbinary:
### Why
As i was working through rubocop offenses, [`Security/YAML.load`](https://docs.rubocop.org/rubocop/cops_security.html#securityyamlload) surfaced in several places. While the easy fix for this is to use `.safe_load` instead, i was also curious why we were using YAML in the first place.

It appears we are using `YAML.load` as a way to convert a string to a boolean. While this works, i don't find it to be intention revealing. This PR introduces an alternate way to solve this.

### What
- [x] Introduces a `ToBoolean` [refinement](https://devdocs.io/ruby~2.7/syntax/refinements_rdoc) allowing us to  cast strings and symbols by calling `.to_boolean`.
- [x] Refactor the `yes_no` helper to accept either a string or a boolean and use `.to_boolean` internally.
- [x] Refactor `SubmissionResponse#name` which i believe does not need to be coercing `boolean_response`.
- [x] Refactor `_submission_responses.html.erb` to use `context.can_admin?` instead of `YAML.load`.
- [x] Refactor all other uses of `YAML.load` to use the new refinement.

### API design
I considered three alternative APIs:
1. A utility method such as `Util.to_boolean`. I usaully prefer the explicitness of this approach over modifying ruby objects, but in this instance, `.to_boolean` felt like a natural additional, similar to other casting methods such as `.to_i`, `to_f` etc.
2. Monkey-patch, which would lead to the same `.to_boolean` interface, but without the need to invoke `using ToBoolean`. This is a more well-known approach but comes with all the issues of monkey-patching, such as lack of transparency.
3. Monkey-patch with a refinement, which has the benefits of better isolation/encapsulation, explicitness and transparency, at the cost of a little more boilerplate. I also found another drawback: `using` doesn't work in erb templates; however this can be circumvented by pushing logic into helpers or presenters (as we do here in the `yes_no` helper).

### `.to_boolean` implementation
This implementation recognizes truthy strings from the set of the following:
* Values recognized as true by [YAML.load](https://yaml.org/type/bool.html).
* The *opposites* of values in `ActiveRecord::Type::Boolean::FALSE_VALUES`.
* Stripped, case-insensitive versions of both the above.

Importantly, there are functional differences from both YAML.load and ActiveRecord. This implementation is an accept-list of truthy strings, making all other strings false (except "" and nil, which return nil).

This is the opposite of AR casting, which is an accept-list of _falsey_ strings, and other strings are considered truthy. I feel it's usually more important to strictly recognize _truthy_ strings than falsey strings. Eg,
`params[:admin].to_boolean` should only return true when it exactly matches a known true value.

YAML.load has accept-lists for both truthy and falsey values, returning a string for anything unrecognized, so this implementation differs in defaulting unrecognized strings to false.

I think these differences are to be expected. The conversion from string to boolean is very contextual and i suspect this is why there isn't a standard `.to_b` implemented in ruby. I'm hoping the rules chosen here are appropriate for our common use cases.

### Testing
Added:
- [x] `spec/refinements/to_boolean_spec.rb`
- [x] `spec/models/submission_response_spec.rb`

### Next Steps
- [ ] There may be other places we can use `.to_boolean`. I've only converted places that were using `YAML.load`

* See inline comments for some specific questions
* ?

### Security
I believe the pre-existing use of `YAML.load` was a significant attack vector because we were using it to convert user input in several places. See [this article](http://tenderlovemaking.com/2013/02/06/yaml-f7u12.html) for details.

### Pre-Merge Checklist

- [x] Any next steps have been turned into Issues or Discussions as appropriate

### DIFF
--- a/app/controllers/matches_controller.rb
+++ b/app/controllers/matches_controller.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+using ToBoolean
+
 class MatchesController < AdminController
   before_action :set_match, only: %i[edit update destroy]
 
@@ -104,7 +106,7 @@ class MatchesController < AdminController
 
     @communication_logs = CommunicationLog.where(match: @match)
 
-    @edit_connection_mode = YAML.load(params[:edit_connection_mode].to_s)
+    @edit_connection_mode = params[:edit_connection_mode].to_boolean
   end
 
   def match_params
--- a/app/models/submission_response.rb
+++ b/app/models/submission_response.rb
@@ -6,21 +6,15 @@ class SubmissionResponse < ApplicationRecord
 
   def name
     # self.public_send(Question::INPUT_TYPES_AND_STORAGE[question.input_type])
-    name = nil
     [
       string_response,
       text_response,
-      YAML.load(boolean_response.to_s),
+      boolean_response.to_s,
       integer_response.to_s,
-      array_response.join(', '),
+      array_response&.join(', '),
       date_response&.strftime('%Y-%m-%d'),
-      datetime_response&.strftime('%Y-%m-%d @ %l:%m')
-    ].each do |string|
-      if string.present?
-        name = string # FIXME: this will find the _last_ present string, is that what we want?
-      end
-    end
-    name
+      datetime_response&.strftime('%Y-%m-%d @%l:%M%P')
+    ].find(&:present?)
   end
 
   # def input_type
--- a/app/refinements/to_boolean.rb
+++ b/app/refinements/to_boolean.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+# Provides a way to cast to boolean
+#
+# Returns true if given any string or symbol that, stripped and downcased, matches:
+#   1. True values recognized by YAML.load: %w[y, yes, true, on] (https://yaml.org/type/bool.html)
+#   2. The opposites of ActiveRecord::Type::Boolean::FALSE_VALUES -- %w[false, 0, f, off]
+# Returns nil if given nil or a blank string
+# Returns self if given a boolean
+# Returns false given any other string or symbol
+#
+module ToBoolean
+  refine String do
+    def to_boolean
+      blank? ? nil : strip.downcase.in?(%w[true yes on t y 1])
+    end
+  end
+
+  refine Symbol do
+    def to_boolean
+      to_s.to_boolean
+    end
+  end
+
+  refine(NilClass)   { def to_boolean; self end }
+  refine(TrueClass)  { def to_boolean; self end }
+  refine(FalseClass) { def to_boolean; self end }
+end
--- a/app/views/submissions/_submission_responses.html.erb
+++ b/app/views/submissions/_submission_responses.html.erb
@@ -5,7 +5,7 @@
   <% r = response %>
   <% question = r.custom_form_question %>
 
-  <% if params && YAML.load(params[:admin].to_s) %>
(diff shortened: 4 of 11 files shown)


app/controllers/matches_controller.rb
```diff
@@ -105,7 +107,7 @@ def set_form_dropdowns
 
       @communication_logs = CommunicationLog.where(match: @match)
 
-      @edit_connection_mode = YAML.load(params[:edit_connection_mode].to_s)
+      @edit_connection_mode = params[:edit_connection_mode].to_boolean
```
exbinary, line 110: There are a number of places where the existing implementation called `.to_s`. I think this was unnecessary, but i might be missing something?

app/models/submission_response.rb
```diff
@@ -6,14 +6,15 @@ class SubmissionResponse < ApplicationRecord
 
   def name
     # self.public_send(Question::INPUT_TYPES_AND_STORAGE[question.input_type])
-    name = nil
-    [string_response, text_response, YAML.load(boolean_response.to_s), integer_response.to_s,
```
exbinary, line 10: I believe `YAML.load(boolean_response.to_s)` was equivalent to `boolean_response`. The changes below are mostly a refactoring to simplify but also drops the unnecessary double-conversion.

app/refinements/to_boolean.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+# Provides a way to cast to boolean
+#
+# Returns true if given any string or symbol that, stripped and downcased, matches:
+#   1. True values recognized by YAML.load: %w[y, yes, true, on] (https://yaml.org/type/bool.html)
+#   2. The opposites of ActiveRecord::Type::Boolean::FALSE_VALUES -- %w[false, 0, f, off]
+# Returns nil if given nil or a blank string
```
exbinary, line 8: Curious if anyone has thoughts about returning `nil` given a `nil` or blank?

app/views/submissions/_submission_responses.html.erb
```diff
@@ -5,7 +5,7 @@
   <% r = response %>
   <% question = r.custom_form_question %>
 
-  <% if params && YAML.load(params[:admin].to_s) %>
+  <% if context.can_admin? %>
```
exbinary, line 8: I believe this was the intention of the existing code?

app/refinements/to_boolean.rb
h-m-m: Since we're using this in the context of parsing form parameters, I also suspect this is the behavior we want.

app/views/submissions/_submission_responses.html.erb
h-m-m: Either way, I appreciate getting the conditionals moved out of a view

app/controllers/matches_controller.rb
h-m-m: I think if we do a sanity check by stepping through it in multiple browsers, that's enough for me to make sure

exbinary:
@h-m-m , @svileshina , i forget where we landed with this PR. Is there anything outstanding? If not could one of you approve please? Thanks!

h-m-m (approved):
I like this improvement.

We've talked about the tradeoffs of this outside this PR, about going this route versus the potential of using a utility method instead, and weren't able to agree whether one is unquestionably better than the other

This code is both written and works, which is more I can say for any other approach we've discused. Looks good to me

PR #948, opened 2021-04-21
Rubocop linting and performance violations

exbinary:
## Why
Building on #935, fixes violations of `Lint/*` and `Performance/*` rules.

### What
See individual commits for the list of rule violations addressed. Diffs probably also make more sense one commit at a time.

In many instances, i left `rubocop:todo`s instead of fixing the violation (this is like `rubocop:disable` but reveals that we really want to fix the violation instead of disable it).

See inline PR comments for specific questions.

### Testing

### Next Steps
?

?

### Pre-Merge Checklist


app/controllers/contributions_controller.rb
```diff
@@ -52,9 +52,9 @@ def contribution_blueprint_options
   end
 
   def filter_params
-    return {} unless allowed_params&.to_h.any?
+    return {} unless allowed_params.to_h.any?
```
exbinary, line 55: I don't think `allowed_params` can be nil?

app/forms/base_form.rb
```diff
@@ -20,20 +20,6 @@ def self.filter_keys(source = self)
...
-    # FIXME: ActiveInteraction::Base#given? seems to have a bug recognizing multipart date params
-    raw_inputs = @_interaction_inputs
-    inputs.select do |key|
-      given?(key) || (
-        raw_inputs.key?(:"#{key}(1i)") &&
-        raw_inputs.key?(:"#{key}(2i)") &&
-        raw_inputs.key?(:"#{key}(3i)")
-      )
-    end
-  end
```
exbinary, line 33: Moved to the end of the file.

config/initializers/simple_form_bulma.rb
```diff
@@ -543,56 +543,56 @@
     # end
 
 # rubocop:enable Layout/CommentIndentation
+end
```
exbinary, line 546: Everything below here in this file is just whitespace change; moved the class definitions below out of the block above.

db/migrate/20200523184552_add_auto_generated_to_communication_log.rb
```diff
@@ -1,7 +1,7 @@
 class AddAutoGeneratedToCommunicationLog < ActiveRecord::Migration[6.0]
   def up
     add_column :communication_logs, :auto_generated, :boolean
-    change_column_default :communication_logs, :auto_generated, :false
+    change_column_default :communication_logs, :auto_generated, false
```
exbinary, line 4: Fixing but `:false` seems to set the default correctly.

db/scripts/submission_response_seeds_gitignored_csvs.rb
```diff
@@ -7,13 +7,14 @@
 importer = Importers::SubmissionResponseImporter.new(User.first, csv)
 importer.import(path)
 
+# TODO process offers and old_offers as well
```
exbinary, line 10: I've tried to interpret the meaning of the pre-existing comment here.

app/controllers/contributions_controller.rb
h-m-m: Right. It can be empty/blank or it can raise an error, but it can't be `nil`

exbinary:
Reviewed in person with @svileshina .

PR #917, opened 2021-03-26
Jade dickinson show contributor name

svileshina:
### Why
This adds the contributor's name to a contribution -  Working off of https://github.com/rubyforgood/mutual-aid/pull/791

### What
Added the contributor name to contributions through contribution.person, so added person to the contribution blueprint and appropriate tests and supporting files 

### Next Steps
- [ ] Add a conditional so this is only displayed for logged in users with admin permissions #933
- [ ] Add the same functionality to Map View #934 

### Outstanding Questions, Concerns and Other Notes
We would like to replace raw json (i.e. lib/contributions.json) in the future

### Meta
We added Jade as a contributor because this was based on their work (https://github.com/rubyforgood/mutual-aid/pull/791) and closed their original PR that had merge conflicts 

### Pre-Merge Checklist

- [x] Security & accessibility have been considered
- [x] Tests have been added, or an explanation has been given why the features cannot be tested
- [x] Documentation and comments have been added to the codebase where required
- [ ] Entry added to CHANGELOG.md if appropriate
- [ ] Outstanding questions and concerns have been resolved
- [ ] Any next steps have been turned into Issues or Discussions as appropriate

PR #950, opened 2021-04-24
Spike use of presenters during serialization

exbinary:
### Why
This is a spike on top of #917 (and #951, #952) experimenting with using Presenters to solve the problem of conditionally showing attributes based on authorization policies.

### What, How
- [x] Add a `BasePresenter` with some functionality useful across presenters
- [x] Add a `present` helper method for use in erb templates
- [x] Fix a bug in `PersonPolicy` that allowed guest users access to any other Person w/o an associated user
- [x] Simplify the display of `person.name` on `/contributions` by not trying to split the name
- [x] Add `ContributionPresenter` and thread it into the serialization process in `ContributionsController#index`
- [x] Add `PersonPresenter`, use it in `ContributionPresenter` and have it conditionally show the name based on `PersonPolicy`.

Probably easiest to review each commit individually.
### Testing
Adds some specs for new code.

### Next Steps
- [ ] See thoughts below about how to support several conditional attribute.
- [ ] ?

This solution works and feels like the right place to put logic.

However i wonder how it would scale when we need to conditionally show more than one attribute. Feels like we may need something analogous to `permitted_attributes` which lists all attributes allowed to be submitted, but in this case, a listing of attributes allowed to be shown. Maybe a `visible_attributes` method on the policy that could be used by the presenter to dynamically mask any disallowed attributes?

### Pre-Merge Checklist

### DIFF
--- a/app/blueprints/contribution_blueprint.rb
+++ b/app/blueprints/contribution_blueprint.rb
@@ -16,14 +16,10 @@ class ContributionBlueprint < Blueprinter::Base
     contribution.created_at.to_f * 1000 # Javascript wants miliseconds, not seconds
   end
   field :type, name: :contribution_type
-  field :profile_path do |contribution, options|
-    options[:profile_path]&.call(contribution.person_id)
-  end
-  field :view_path do |contribution, options|
-    options[:view_path]&.call(contribution.id)
-  end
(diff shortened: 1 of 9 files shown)


exbinary:
Closing in favor of #953 .

PR #951, opened 2021-04-28
Fix bug in PersonPolicy that allows guest to access other People

exbinary:
### Why
Discovered a bug while working on #950, extracted the fix here so it can be merged post haste.

### What
`PersonPolicy.own_person` was implemented as `person.user_id == user&.id`.
With a guest user, where user is nil, this works but only if person has a user; if it doesn't, then the two nils match and `own_person` returns true.

This effectively meant any guest user could access any Person record that didn't have an associated user record.
This is actually common because listings can be created by guests, in which case we generate a Person record w/o a user to associate it with.

To recreate, guest users need to guess the record id of a an existing `Person` record with no associated `User`.

### Testing
- [x] Added specs to cover this case in `person_policy_spec.rb`

### Next Steps
?

?

### Security
I believe this fixes an existing vulnerability.

### Pre-Merge Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,13 @@ We're still in the [initial development phase](https://www.jering.tech/articles/
 This changelog also serves to acknowledge the incredible people who've contributed brilliance, effort and being. Their handles are listed under the first release they each  touched. 💗🙏🏾
 
 ## [Unreleased]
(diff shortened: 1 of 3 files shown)


h-m-m (approved):
Nice catch

maebeale (approved):

PR #952, opened 2021-04-28
Presenter improvements

exbinary:
### Why
Extracts some small improvements to our presenter code out of #950 

### What
- [x] Extract a `BasePresenter` providing convenience accessors and a default initializer
- [x] Add a `present` helper method that handles wiring `ActionView::Contex` into presenters
- [x] Refactor `announcements/show` to not neet the `_presenter` suffix

?

### Pre-Merge Checklist
- [x] Security & accessibility are not impacted

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -87,4 +87,9 @@ module ApplicationHelper
     # There should always be a current org, but being defensive here helps simplify tests
     context.host_organization&.logo_url.presence || asset_pack_path('media/images/logo.png')
   end
+
+  def present object, presenter_class
+    presenter = presenter_class.new object, self
+    block_given? ? yield(presenter) : presenter
+  end
 end
--- a/app/presenters/announcement_presenter.rb
+++ b/app/presenters/announcement_presenter.rb
@@ -1,4 +1,4 @@
-class AnnouncementPresenter < SimpleDelegator
+class AnnouncementPresenter < BasePresenter
   def expires_on
     return "This announcement does not expire." unless publish_until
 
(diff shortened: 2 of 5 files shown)


h-m-m:
This is a fascinating convention. Have you seen or used this in other code bases?

h-m-m:
Also, would we use nested blocks if we wanted a view that used two presenters?

I also think of presenters ideally being for a specific context. Using this announcement example, we might want to present info differently on an admin page than in a public-facing page. If the info isn't context-specific, would it be better in the model itself or, if we want super thin models, in some class that is the default interface to the model that then itself gets presented? (Would that second option be essentially a presenter anyway? Would that second option be too much abstraction for our needs so far?)

exbinary:
> This is a fascinating convention. Have you seen or used this in other code bases?

Which convention are you referring to? :)

exbinary:
> Also, would we use nested blocks if we wanted a view that used two presenters?

I think nested blocks would work fine, though `present` also returns the presenter; the view can always choose to assign both presenters to variables instead of using the block form.

exbinary:
> I also think of presenters ideally being for a specific context. Using this announcement example, we might want to present info differently on an admin page than in a public-facing page. If the info isn't context-specific, would it be better in the model itself or, if we want super thin models, in some class that is the default interface to the model that then itself gets presented? (Would that second option be essentially a presenter anyway? Would that second option be too much abstraction for our needs so far?)

Hmm, great questions! As we've discussed before, i'm definitely in favor of context specific presenters instead of the 1 model to 1 presenter norm that's common. I like all the solutions you listed and think we can use them in combination as appropriate. IE:
1. push things into the model that really belong there
2. start with context specific presenters
3. when we see significant duplication between presenters, look to composition

PR #953, opened 2021-04-29
Custom Blueprinter Transformer to hide restricted fields

exbinary:
### Why
As an alternative to the spike in #950, this solution uses a custom blueprinter transformer as a way to hide
restricted attributes based on pundit policies.

### How
[Blueprinter](https://github.com/procore/blueprinter) supports custom transformers (see documentation in the 'Transform Classes' section). Here, we use that feature to
- [x] create a `HideRestrictedFieldsTransformer` which, when specified in a serializer, looks up a pundit `Policy` for the object being serialized and uses that to hide restricted fields, which are
- [x] specified by a new custom method an policies named `restricted_attributes`. This method mirrors the canonical `permitted_attributes`. The latter restricts attributes being _submitted_ while the latter restricts attributes being _serialized_.

### Testing
Added an initial spec to smoke test the solution but probably need a few more to see how associations works

### Next Steps
- [x] Add specs for `association` fields and `views`.
- [ ] Add documentation for this custom pattern.
- [ ] Create an issue on `blueprinter` about transforms not being applied to views automatically.

My initial reaction is that this feels cleaner than the presenter solution in #950.

Comparative Negatives:
* It is a custom hook, which  certainly incurs a long term maintenance cost
* There's some 'convention magic' involved (we have to know/document that the transform will look for a `restricted_attributes` method, and that it has to be explicitly invoked in the blueprint).

Positives:
* The overall boilerplate and impact on the codebase is much less than having to create presenters for every object being serialized.
* Is a cleaner, simpler solution than having to override every restricted attribute individually in the presenter.
* Is likely more performant that decorating each object.

### Pre-Merge Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -68,7 +68,7 @@ GEM
     bindex (0.8.1)
     binding_of_caller (0.8.0)
       debug_inspector (>= 0.0.1)
-    blueprinter (0.23.4)
+    blueprinter (0.25.3)
     bootsnap (1.4.6)
       msgpack (~> 1.0)
     builder (3.2.4)
--- a/app/blueprints/blueprint_authorization.rb
+++ b/app/blueprints/blueprint_authorization.rb
@@ -1,0 +1,14 @@
+module BlueprintAuthorization
+  def self.extended(blueprint)
+    blueprint.class_eval do
+      transform HideRestrictedFieldsTransformer
+    end
+  end
+
+  def view(*options, &block)
+    super(*options) do
+      transform HideRestrictedFieldsTransformer
+      block.call
+    end
+  end
+end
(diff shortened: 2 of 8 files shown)


exbinary:
Update @maebeale, @svileshina, @h-m-m : while adding specs for associations and views, ran into one of those edges we were concerned about with this approach. Turns out that specifying a `transform` in a blueprint doesn't propagate to its `view`s, meaning we'd have to explicitly specify the transform in each separate view (as show in 6fb7742).

That didn't feel serviceable to me so i've refactored to apply the transformer by default to all blueprints, and have it no-op if the policy doesn't implement `restricted_attributes`. See ba58552 for the significant changes and please lmk what you think of this change to what we discussed.

svileshina (approved):

exbinary:
@h-m-m , @svileshina: i've refactored based on Howard's idea of overriding blueprinter's `view` macro to ensure it applies the transformer to views as well as the root blueprint. Please give the updated code another look.

PR #956, opened 2021-05-08
Lay groundwork for system specs

exbinary:
### Why
We've been needing some system-level test coverage for a while now. This PR sets up the minimal groundwork needed to start writing them.

### What
- [x] Add necessary gems
- [x] Add first, nominal system spec
- [x] Configure rspec to not run system specs with non-system specs. Eg, invoking `bin/rspec` will exclude all system specs; to run them, all requested specs must be system specs, eg `bin/rspec spec/system`. 
- [x] Configure CirclCI to run system specs as a separate job
- [x] Provide ability to run specs with 'headless_chrome' based on an env. variable.

### Next Steps
With these changes, our `.circleci/config.yml` is in need of some refactoring; will follow up with a PR to address this.

### Outstanding Questions, Concerns and Other Notes
?

### Pre-Merge Checklist
- [x] Security & accessibility are not impacted
- [x] Tests have been added, or an explanation has been given why the features cannot be tested
- [ ] Documentation and comments have been added to the codebase where required
- [ ] Entry added to CHANGELOG.md if appropriate
- [ ] Outstanding questions and concerns have been resolved
- [ ] Any next steps have been turned into Issues or Discussions as appropriate

PR #957, opened 2021-05-08
CircleCI config refactorings

exbinary:
### Why
With the addition of system specs in #956, some opportunities to refactor our CircleCI config have become evident.

### What
I recommend reviewing each commit separately, in order:
- [x] Introduce [executors](https://circleci.com/docs/2.0/configuration-reference/?section=configuration#executors-requires-version-21) to dry up docker images along with their base environment variables.
- [x] Introduce use of the [workspace](https://circleci.com/docs/2.0/concepts/?section=getting-started#caches-workspaces-and-artifacts) to cache directories within workflows (in contrast to 'caches' which are useful _between_ workflows).
- [x] Extract a new `webpacker` job and corresponding `webpacker_compile` command
- [x] Cache the results of `webpacker:compile` to improve build speed where possible

### Testing
Tested by pushing this branch to CI.

### Next Steps
- [ ] Would like to discuss the `webpacker:compile` caching strategy.
- [ ] Read up on what the shell `set` and `pipefail` options do and evaluate whether they are needed in all the places they're being used.
- [ ] Look for other instances of hidden coupling between and amongst jobs and commands (eg: the rspec output file location was specified in the rspec command and also the calling job).
- [ ] Any further refactorings?

* What belongs in 'commands' vs. 'jobs'? Eg, does caching belong in a command or in the job that calls it?
* Should we come up with a consistent naming convention to disambiguate similarly named workflows, jobs and commands?

### Meta
Up until now, i hadn't taken the time to really read up on and understand the components that make up a CCI config file. This has been fun to learn and makes a lot more sense now!

### Pre-Merge Checklist
- [x] Security & accessibility are not impacted

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,14 @@ commands:
           key: node-cache-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}
           paths:
             - node_modules
+  webpacker_compile:
+    steps:
+      - run:
+          name: 'Compile webpacker assets'
+          command: |
+            set -euxo pipefail
+            bundle config path .bundle
+            bundle exec rails webpacker:compile
   run_rubocop:
     description: 'Run rubocop'
     steps:
@@ -66,7 +74,6 @@ commands:
             set -euxo pipefail
             bundle config path .bundle
             bundle exec rails db:create
-            bundle exec rails webpacker:compile
             bundle exec rspec --format progress --format RspecJunitFormatter -o /tmp/rspec_system/rspec.xml spec/system
 
 executors:
@@ -102,6 +109,17 @@ jobs:
       - persist_to_workspace:
           root: .
           paths: [node_modules]
+  webpacker:
+    executor: app_only
+    environment:
+      RAILS_ENV: test
+    steps:
+      - checkout
+      - attach_workspace: {at: .}
+      - webpacker_compile
+      - persist_to_workspace:
+          root: .
+          paths: [public/packs-test]
   rubocop:
     executor: app_only
     steps:
@@ -149,7 +167,12 @@ workflows:
       - frontend_specs:
           requires:
             - yarn
+      - webpacker:
+          requires:
+            - bundle
+            - yarn
       - system_specs:
           requires:
             - bundle
             - yarn
+            - webpacker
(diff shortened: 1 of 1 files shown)


h-m-m:
> Read up on what the shell set and pipefail options do and evaluate whether they are needed in all the places they're being used.

Here's a writeup about it https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

.circleci/config.yml
```diff
@@ -30,130 +30,133 @@ commands:
...
+            - webpacker-cache-{{ .Branch }}
+            - webpacker-cache-main
+      - run:
+          name: 'Compile webpacker assets'
+          command: |
+            set -euxo pipefail
+            bundle config path .bundle
+            bundle exec rails webpacker:compile
+      - save_cache:
+          key: webpacker-cache-{{ .Branch }}
```
h-m-m, line 46: In our meeting today @exbinary brought up this problem, where caches are write-once:

Screen Shot 2021-05-08 at 11 08 21 AM

I dug into this and ran a few test runs on a separate branch.

It looks to me like the problem here is that webpack is putting everything in one unitary file (one file each for css & js). It doesn't split anything up and doesn't persist any useful intermediary steps to disk that we can then cache.

We can configure webpack differently to get something cacheable<sup>[1]</sup>. I'm not sure how long that would take, though. As a theoretical example, it should be possible to have one file that's all the packages we're pulling in and another that's only the code we've written that then pulls in the other<sup>[2]</sup>. I'm worried how hard it might be to get Rails to play nice with this, though<sup>[3]</sup>.

[1]: https://webpack.js.org/guides/code-splitting/#
[2]: https://dev.to/nejremeslnici/split-your-webpacker-bundles-to-speed-up-the-web-2
[3]: https://dev.to/rossta/these-rails-apps-are-overpacking-their-javascript-bundles-3gl5

h-m-m:
I like looking at the chart at https://xkcd.com/1205/ sometimes when I'm figuring how much time to spend on something. Here's some rough guess math:

We had ~50 runs in the past week. I'll round that down to 5/day with the understanding that not every run is one where you're waiting for the output before you can move on. This improvement could save us 1 minute per run if it works, so the chart says that ~6 days of one person's effort on this will break even in 5 years.

If we further give ourselves a 50/50 chance of actually getting it to work, we might want to spend 3 days on it and then reassess

Issue #433, opened 2020-05-27
Make admin section of Navbar.vue be a dropdown


maebeale:
- (I think we want to) keep Matches as a navbar item
- Keep Feedback as a button, or move to menu?
- Move Logout under Account
- Add button to My Profile (where there's a button to edit Account -- is that too buried?)

Issue #365, opened 2020-05-18
User is able to get to Edit account from nav


maebeale:

armahillo:
Assuming we're not doing anything weird with devise, this is normally straighforward. I've not done much with Vue and haven't reviewed the app enough to know, but if it's the standard approach I'll put up a PR and retroactively self-assign

maebeale:
AWESOME! Yeah, you can ignore the vue stuff. Totally straight devise.
Changing password is SO helpful!!!

(We also can't currently create Users bc of a devise reroute.)

armahillo:
Is there an existing "profile" page for users to manage their own profile? 

I can get to `http://localhost:3000/users/edit?locale=en` via the Admin area

maebeale:
Yeah, I took it out of the nav bc we have so many things. Was going to make
a dropdown under Logout to include it there, but don't know vue well enough
yet to pull that off. Figured if needed, for now, ppl could go to the Users
index and click edit button.

armahillo:
So it appears this feature already exists, it just needs to be linked somewhere.

I think this issue should probably be re-written to be "Add link to header for /users/edit"? There will need to be authorization set up for this action, eventually, so that people don't URL inject other user's edit pages, something like:

```ruby
def edit
    if @user == current_user || !current_user.is_admin?
      redirect_to edit_user_registration_path
    else
      set_form_dropdowns
    end
  end
```

armahillo:
Just confirmed I can change my password via `/users/edit` -- so really we just need a link added to the nav, and (eventually) access-control set

maebeale:
Wuttt???? Um, ok then!

maebeale:
Is the change my password button working when not logged in? Maybe it was meant about that?

armahillo:
I just verified:
> As a signed-out user, when I attempt to access /users/edit and /users/1/edit, I am sent to a sign-in page

We should probably create some system specs for this stuff.

maebeale:
YESSSS

armahillo:
I dont see any system specs right now and that requires enough configuration change I think it should probably be its own issue

h-m-m:
Is there any of this that is dependent on getting Pundit deployed first?

exbinary:
No longer an issue.

PR #963, opened 2021-05-20
433-365 : dropdown on navbar and profile

vurtn:
Sorry for my english.

### Why
The main issue i tried to solve (#433) is about adding a dropdown in the right part of the navbar. I added a link to edit the profile of the current user (#365). 

### What
- [x] Add a dropdown menu in the navbar
- [x] Create some items in it
- [x] Change the z-index of the navbar
- [x] Add a new component `DeleteLink`
- [x] Change the `FeedbackButton`
- [x] Change some things in `NavBar.spec.js`
- [x] Add a test in `user_spec.rb`

### How
I added the dropdown in the navbar with Buefy and had to make some changes.

First was the buttons/button style in the navbar. I added a parameter to `FeedbackButton` because it is already a link and the value by default is the button style (it's keeping the current style by default to not break some code elsewhere (and it's name is Button so... :smile: ). The button to disconnect the user is a real button so i created another component to have the same behavior but with a link.  

I used the Logout condition to show the dropdown in the navbar and added a z-index on the navbar because some components are above sometimes. It's visible on small screens like on this gif :
nav-zindex
The burger on the page is above the navbar.

In the dropdown menu there is a link to edit our profile :
nav-profile

### Testing
Some test failed due to the new structure of the navbar in `NavBar.spec.js`. The main issue was this part :

```
const navLinks = $wrapper.findAll('a.navbar-item').wrappers
const navButtons = $wrapper.findAll('.navbar-item a').wrappers
const navForms = $wrapper.findAll('.navbar-item form').wrappers
return [...navLinks, ...navButtons, ...navForms].map(link => link.text())
```

The dropdown div is a `navbar-item`, so the links inside it match `a.navbar-item` and `.navbar-item a` selectors. The return is an array and not a set so i had duplicates and the test failed. I changed the lines above to :

```
const navLinks = $wrapper.findAll('a:not(.burger)').wrappers
return [...navLinks].map((link) => link.text())
```
I kept the spread operator because if we need to select something else, we will just add the new catches to this array like before. The navbar contains only links now so the only duplicate i had was the root page because of the burger link for the small screens. I excluded it of the query selector. I added the new links to the test of admin buttons. 

I added a test in `user_spec.rb` to verify that the user is redirected to the sign in page if he is not logged in while trying to access the profile page.

### Next Steps

Unknown. Maybe create some tests with a context like "with the right credentials" (and its complementary of course) to check the access to resources.

Webpack throwing this error sometimes when i'm testing and i still don't understand why : 

```
ActionView::Template::Error:
       Webpacker can't find application.css in /my-path/mutual-aid/public/packs-test/manifest.json.   
``` 

### Accessibility
Not seem to be concerned now.

### Security
Not seem to be concerned (now ?). 

### Meta
I didn't search why but i created a new account in the app and have the volonteer profile. The `FeedbackButton` is shown in the navbar but when i click on it i have this message : 

> Sorry, you aren't authorized to do that.
> If you think this is wrong, please contact your administrator.

I think something is wrong :smile: The `FeedbackButton` have the `v-if="visible('Feedback')"` condition and the item is shown but i am not authorized to access it. 

### Pre-Merge Checklist
- [ ] Any next steps have been turned into Issues or Discussions as `appropriate`

Best regards and stay safe !


app/assets/stylesheets/application.scss
```diff
@@ -4,20 +4,24 @@
...
   text-align: center;
 }
 
 .section-detail {
   padding: 0rem 1.5rem 3rem 0rem;
 }
+
+.navbar {
+  z-index: 10;
+}
```
exbinary, line 27: I wasn't able to recreate the problem you mentioned with the 'burger' showing up in the wrong place. When i comment out this z-index, everything still seems to work fine. Could you double check and describe steps to recreate if you're still seeing it happen?

app/javascript/components/NavBar.vue
```diff
@@ -13,39 +13,51 @@
...
       </b-navbar-item>
+
+      <b-navbar-dropdown label="Account" collapsible right v-if="visible('Logout')">
+        <div class="sub-item">
+          <a class="navbar-item" href="/users/edit">My profile</a>
+        </div>
+        <div class="sub-item">
+          <FeedbackButton
+            classes="navbar-item"
+            action="software_feedbacks/new"
```
exbinary, line 38: This `action` property seems to be obsolete (i know it existed before your changes -- must've been inadvertently left behind previously).

app/javascript/components/forms/DeleteLink.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <form :action="action" method="post" ref="form">
+    <AuthTokenInput />
+    <input type="hidden" name="_method" value="delete" />
+    <a href="#" :class="classes" v-on:click="submit"><slot /></a>
```
exbinary, line 5: Minor: i have a slight preference to put `<slot />` on its own line, to give it a little more visibility.
When scanning a component template, it's helpful to be able to quickly spot any slots defined.

exbinary (approved):
@vurtn , i went ahead and made the minor changes from the previous review.
Thanks so much for your work on this! 🙌🏾 .

Issue #483, opened 2020-06-09
Replace is-grouped with is-inline-flex on forms


maebeale:

gamecat2d:
i will give it a try !

h-m-m:
Resolved. Thanks @finn2d !

PR #964, opened 2021-05-22
483 : replace is-grouped by is-inline-flex

finn2d:
Sorry for my english.

### Why
Replace the class is-grouped by is-inline-flex (#483) 

### What
- [x] Replace is-grouped by is-inline-flex in views

### How
By searching the keyword in the project. :sweat_smile: 

### Testing
I don't know what to add, the tests are still full green.

### Next Steps
Unknown.

Not seem to be concerned.

### Accessibility
Not seem to be concerned.

### Security
Not seem to be concerned.

### Meta
Not seem to be concerned.

### Pre-Merge Checklist

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -4,16 +4,16 @@
 @import './custom_spacing_helpers';
 @import './bulma_simple_form_customization';
 
-@import './yearbook';  // TODO: remove this once listings/new is refactored
-@import './listings';  // TODO: remove this once listings/new is refactored
+@import './yearbook'; // TODO: remove this once listings/new is refactored
+@import './listings'; // TODO: remove this once listings/new is refactored
 
 @import './custom_text';
 
 @import './mapbox-gl';
 @import './maps';
-@import "./actiontext.scss";
+@import './actiontext.scss';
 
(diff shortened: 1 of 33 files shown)


h-m-m:
This change looks great. I assume there's some autoformatter that cleaned up the CSS here, and I like those cleanups, too.



For our own future reference since some context is missing from the issue, it looks like we gain two things from this change.

First, `is-grouped` comes from Buefy so it really should only be used for Vue components. Bulma provides `is-inline-flex`, so it's fine for use site-wide.
Second, `is-inline-flex` looks better at smaller widths.

Screen shots from playing around with this locally:

#### Before 
(with is-grouped)
Wider breakpoint:
Screen Shot 2021-06-26 at 10 23 51 AM
Smaller breakpoint:
Screen Shot 2021-06-26 at 10 26 07 AM

#### After
(with is-inline-flex)
Wider breakpoint:
Screen Shot 2021-06-26 at 10 24 27 AM

Smaller breakpoint:
Screen Shot 2021-06-26 at 10 26 27 AM

So at the wider breakpoint they look identical, and at the narrower width the `is-inline-flex` leads to less need for left-right scrolling

h-m-m (approved):

Issue #827, opened 2021-01-18
Validate Emails


svileshina:
We need to validate emails on Person (i.e. john@example shouldn't be possible to add)

gamecat2d:
i give it a try !

exbinary:
Fixed in #965

PR #965, opened 2021-05-22
827 : emails validation in person

gamecat2d:
Sorry for my english.

### Why
We need to validate emails in person (#827).

### What
- [x] Create a validator
- [x] Add a test in `person_spec.rb`

### How
The validator is in the guide about [validations](https://guides.rubyonrails.org/active_record_validations.html#custom-validators).

### Testing
A test is available now in `person_spec.rb`.
email-not-valid

### Next Steps
Unknown.

Not seem to be concerned.

### Accessibility
Not seem to be concerned.

### Security
Not seem to be concerned.

### Meta
Not seem to be concerned.

### Pre-Merge Checklist

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -23,8 +23,7 @@ class Person < ApplicationRecord
   accepts_nested_attributes_for :location
 
   validate :preferred_contact_method_present!
-  validates :email, email: true
-  validates :email_2, email: true
+  validates_with EmailValidator, fields: [:email, :email_2]
 
   def name_and_email
     "#{name} (#{email})"
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -23,6 +23,8 @@ class Person < ApplicationRecord
   accepts_nested_attributes_for :location
 
   validate :preferred_contact_method_present!
+  validates :email, email: true
+  validates :email_2, email: true
 
   def name_and_email
     "#{name} (#{email})"
--- a/app/validators/email_validator.rb
+++ b/app/validators/email_validator.rb
@@ -1,8 +1,13 @@
-class EmailValidator < ActiveModel::EachValidator
-  def validate_each(record, attribute, value)
-    return if value.nil? || value.strip.empty?
-    unless /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i.match?(value)
-      record.errors.add attribute, (options[:message] || "is not valid")
-    end
+class EmailValidator < ActiveModel::Validator
+  REGEX_EMAIL = /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i
+  def validate(record)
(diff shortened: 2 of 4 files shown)


app/models/person.rb
```diff
@@ -23,6 +23,8 @@ class Person < ApplicationRecord
   accepts_nested_attributes_for :location
 
   validate :preferred_contact_method_present!
+  validates :email, email: true
+  validates :email_2, email: true
```
h-m-m, line 27: I like the `validates_with` methods a little better, because they use a class name. If the class name `EmailValidator` is here in this validation line, someone reading this code will have to do less thinking to guess where the logic exists. I'm sure there's other ways to write this that accomplish the same thing. What I care about is that these `email: true` arguments feel like too much "Rails magic" too me

spec/models/person_spec.rb
```diff
@@ -24,6 +24,20 @@
...
 
+  describe 'email validation' do
+    let(:contact_method) { build :contact_method, name: 'Email', field: 'email' }
+    subject(:person) { build :person, preferred_contact_method: contact_method, email: 'test@missingtld' }
+
+    context 'when the email field is not valid' do
+      it { is_expected.not_to be_valid }
+
+      it 'generates an error on the correct field' do
+        person.valid?
```
h-m-m, line 35: Do you think these tests would be better as tests on the validator itself instead of on the model?

app/validators/email_validator.rb
```diff
@@ -0,0 +1,8 @@
+class EmailValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return if value.nil? || value.strip.empty?
+    unless /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i.match?(value)
```
h-m-m, line 4: @exbinary went to see if any of our dependencies already have an email regex.

The Devise gem includes an email regex, but I think that one is way too permissive.

@exbinary pointed out that Ruby's standard library includes `URI::MailTo::EMAIL_REGEXP`, which may be good enough for us. What do you think?

h-m-m (commented):
Thanks for writing this PR! I'm very happy to see all the work you've been contributing, and I like your approach to solving this problem

app/validators/email_validator.rb
h-m-m: Oh, never mind — I see that `URI::MailTo::EMAIL_REGEXP` would permit the bad example described in the issue

app/models/person.rb
gamecat2d: Done :+1:

spec/models/person_spec.rb
gamecat2d: Done :+1:

h-m-m (approved):

h-m-m:
I was able to pull this branch to my local dev environment, where it worked with no problems!

Issue #806, opened 2020-11-13
Menu bar not displaying correctly on public contribution forms when the admin option says to


h-m-m:
### Steps to recreate:

- Go to https://mutual-aid-nwmi.herokuapp.com/offers/new https://mutual-aid-nwmi.herokuapp.com/asks/new

additionally, if you have admin access

- Go to the admin settings page (`/system_settings/1/edit?locale=en`) and verify the "display nav bar" setting is on

### Expected results

The nav bar / menu shows up on the form page URLs above

### Observed results

The nav bar is not shown on those pages

gamecat2d:
i give it a try !

exbinary:
I believe this was fixed in #966.

Issue #770, opened 2020-10-07
show_navbar option should toggle display on /community_resources and /announcements


maebeale:
This boolean should only affect:
* community resources
* announcements pages
* asks|offers
* respective confirmation pages for all the above

Whereas right now it seems to affect all (?) pages?

gamecat2d:
Hi, sorry for my english it is not my mother tongue !

I think i found the problem, determine_layout (to hide/show the navbar) is in the controller of community ressources, announcements but in the public pages controller too. It concerns pages like the landing page, about page, version page and maybe contribution soon (?).

Is navbar always shown on the about page, etc ?
If i remove this determine_layout in the public controller and set the display boolean to false, the navbar is always visible except when adding a community ressource or an announcement for example.

Best regards and stay safe !

exbinary:
I believe this was fixed in #966.

PR #966, opened 2021-05-22
806 : the navbar is not displayed correctly

gamecat2d:
Sorry for my english.

### Why
The navbar is always hidden on `asks/new` and `offers/new` (#806).

### What
- [x] Display the navbar if the setting is on 

### How
The problem is that the navigation bar is still hidden because the option in settings is not used. We need to hide the navbar if the boolean is `false` like in other controllers.

### Testing
I don't know what to add.

### Next Steps
Unknown.

Not seem to be concerned.

### Accessibility
Not seem to be concerned.

### Security
Not seem to be concerned.

### Meta
Not seem to be concerned.

### Pre-Merge Checklist

### DIFF
--- a/app/controllers/asks_controller.rb
+++ b/app/controllers/asks_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class AsksController < PublicController
-  layout 'without_navbar', only: %i[new create]
+  layout :determine_layout, only: %i[new create]
 
   def index
     redirect_to contributions_path
@@ -46,5 +46,9 @@ class AsksController < PublicController
     }.to_json
 
     render :new
+  end
+
+  def determine_layout
+    'without_navbar' unless context.system_settings.display_navbar?
   end
 end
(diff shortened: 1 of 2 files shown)


h-m-m:
Hi, @gamecat2d! Thanks for helping us out with this project! I'm so happy to see contributions from new people.

The changes you've made look okay to me. Since you say you're not sure what tests to add, maybe I could help with some suggestions.

Maybe we can add new files to `spec/requests` folder, one for each controller, and then add a test that

- sets the system preference
- requests the page
- asserts that the response.body does or does not have the nav bar

If you haven't written a test like this before, I'm happy to provide examples. Please let me know what you think

h-m-m:
An update, @gamecat2d:

I talked to @exbinary about this, who pointed out that your question here about testing and your questions in #770 raised some interesting issues we hadn't discussed before. We both agree that we should accept this PR and then add some info to a new issue, #967, to clarify requirements for how the nav bar could work

If you want to, feel free to request an invite to our Slack at https://rubyforgood.herokuapp.com/ and join us in the `#mutualaid` channel. We'd be happy to give you more background into what's happening in the code here and answer any questions you have that aren't covered in #967. You're also welcome to continue posting here on GitHub if that's more comfortable for you. Either way is fine — I wanted to have the option.

h-m-m (approved):

Issue #967, opened 2021-05-22
Consolidate navbar display logic


exbinary:
Our current solution for determining whether to display the nav bar is spread across several places:
* layouts:
    * `application.html.erb`
    * `without_navbar.html.erb`
    * `_navbar.html.erb`
    * `_without_navbar.html.erb`
* controllers:
    * Each individual controller where we want to toggle the navbar
* specs:
    * Request specs for each controller

Inspired by @gamecat2d's question about how best to test the changes in #966, @h-m-m and i discussed ways to consolidate our solution in a single place for future maintainability and also be tested in one place.

One potential solution is to:
* consolidate the two layouts into a single one that applies across the whole site
* have it call a helper method that is given the current_controller and:
    * checks whether it is one where we want to toggle the navbar
    * whether to display the navbar
* the testing would then all be consolidated to a unit test of the helper method

gamecat2d:
Sorry for my english. 

I'm new in Ruby (and Rails too) but in the docs i found the [current_page?](https://apidock.com/rails/v6.1.3.1/ActionView/Helpers/UrlHelper/current_page%3F) method.

What do you think about having in memory (or in database to configure it with the back office) the list of controllers/actions where the navbar need to be hidden and reduce it to a boolean value to display or not the menu in the layout ?

Something like (it's not working just an idea) :
```
<% if !hidden_actions.inject(false) { |acc, actions| acc || current_page?(actions)} %>
the navbar
<% end %>
```

Or a hash to be quicker with the controller as a key to know in O(1) the case where the current controller is not concerned because `hidden_actions[current_controller].nil?`

Best regards and stay safe !

h-m-m:
Hi, @gamecat2d!

I'm not sure what will work here, since I haven't tried it myself. In addition to `current_page?` you have some other options available. I think `params[:controller]` and `params[:action]` might work, and there might be `controller_name` and `action_name` methods.

Also, I think @exbinary's suggestion in this issue is to have all the logic and conditionals in a method that's outside of the view. We haven't been testing views directly in this project. Rails makes testing logic in the views difficult. 

For some additional context, when @exbinary and I were talking about this earlier today we were thinking that we'll probably get future requests to change which pages have the nav bar and which pages don't.

Since we probably will get those future requests, I'd prefer a solution where, if I want to make those changes, I only have to change one or two lines of code and also change one test. And since Rails makes testing views difficult, it would be nice if the code I have to change is not in the view

[...]

PR #968, opened 2021-05-23
967: displaying the navbar conditionnaly

gamecat2d:
Sorry for my english.

### Why
We need to have the logic about displaying the navbar in one place and not spread all over the controllers (#967).

### What
- [x] Add a NavbarHelper
- [x] Add a method in application controller that call the NavbarHelper
- [x] Add the call of the method in controller around the navbar in the layout
- [x] Remove the layout without the navbar
- [x] Remove the code in the controllers about the old logic
- [x] Add tests in navbar_helper_spec.rb

### How
I added the function about displaying the navbar in the `application_controller` because with the inheritance, all controllers will have it. Then i created a `NavbarHelper` which have the list of the pages concerned about the option and return `true` or `false` if the navbar can be display.

### Testing
Tests about the navbar helper are in `spec/helpers/navbar_helper_spec.rb`

### Next Steps
I add the current pages (main branch) in the hash and some of them are deleted in others issue (not closed). That's allow us to see clearly what pages had maybe a problem. It should be updated (and close the other issues).

Not seem to be concerned.

### Accessibility
Not seem to be concerned.

### Security
Not seem to be concerned.

### Meta
Not seem to be concerned.

### Pre-Merge Checklist


app/controllers/thank_you_controller.rb
```diff
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class ThankYouController < PublicController
-  layout 'without_navbar'
```
h-m-m, line 4: I always like seeing cleanup like this!

spec/helpers/navbar_helper_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    let(:settings) { true }
+
+    it "should return true for the pages where the navbar can be hidden" do
+      subject.hidden_places.each do |controller, actions|
+        actions.each { |action| expect(subject.show_navbar?(controller, action, settings)).to eq(true) }
+      end
+    end
+
+    it "should return true for the pages not concerned by the displaying option" do
+      expect(subject.show_navbar?('always_display', 'the_navbar', settings))
```
h-m-m, line 12: It might be good here to display data that's more realistic so people can read the test to know what our goals are. For example, setting the first two arguments to `fake_test_controller` and `index` might be a better example. Or maybe `always_displayed_controller`? Whatever you'd like

h-m-m (commented):
Hi, @gamecat2d !

Thanks for helping with this. I feel like you really understood some of the suggestions from https://github.com/rubyforgood/mutual-aid/pull/966, and I appreciate seeing the tests here. I had a few suggestions for your code. Let me know what you think!

app/controllers/application_controller.rb
```diff
@@ -37,10 +38,14 @@ def context
   end
   helper_method :context
 
+  def shown?
+    NavbarHelper.show_navbar?(params[:controller], params[:action], context.system_settings.display_navbar?)
+  end
+
```
exbinary, line 44: I like this direction because it means `show_navbar?` doesn't need to be mixed into every controller.

Pushing that logic a bit further, i don't think we really need the `shown?` method either.

We could instead move this logic directly into `layouts/application.html.erb` since that's the only place it is used:
``` erb
  <%# layouts/application.html.erb %>
  <% if shown_navbar? controller_name, action_name, context.system_settings.display_navbar? %>
    <%= render "layouts/navbar" %>
  <% end %> 
```
This way, nothing extra is added to `ApplicationController`.

Note also the use of `controller_name` and `action_name` above.

app/helpers/navbar_helper.rb
```diff
@@ -0,0 +1,20 @@
+module NavbarHelper
```
exbinary, line 1: The way we're using this module, i don't think it belongs in `app/helpers`, which have a specific usage pattern in rails. 
I'd consider moving this to `app/lib` and naming it `NavbarVisibility` or some such.

app/helpers/navbar_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+  def self.show_navbar?(controller, action, settings)
```
exbinary, line 2: The param name `settings` here is ambiguous. From looking at how it's called, i see it comes from `context.system_settings.display_navbar?`. Perhaps we name it `display_navbar` to match? Open to other suggestions as well.

exbinary (commented):
This is looking great @gamecat2d , much better than the existing solution.
A couple of thoughts below. Thanks!

app/controllers/application_controller.rb
```diff
@@ -37,10 +38,14 @@ def context
...
   private
 
   # TODO: this appears to be unused?
-  def user_not_authenticated(_exception)
+  def user_not_authenticated(exception)
```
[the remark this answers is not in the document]
gamecat2d: Done :+1:

app/helpers/navbar_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+    always_display = hidden_places[controller].nil? || !hidden_places[controller].include?(action)
+    always_display || settings
+  end
+
+  def self.hidden_places
+    hidden_places = {}
+    hidden_places['announcements'] = ['new', 'create']
```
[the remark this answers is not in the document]
gamecat2d: Done with NAVBAR_HIDEABLE :+1:

spec/helpers/navbar_helper_spec.rb
gamecat2d: Done :+1:

app/helpers/navbar_helper.rb
gamecat2d: Done :+1:

app/controllers/application_controller.rb
gamecat2d: Done :+1:

app/helpers/navbar_helper.rb
gamecat2d: Done and now i understand the difference between Helpers directory and lib ! :+1:

app/lib/navbar_visibility.rb
```diff
@@ -0,0 +1,15 @@
...
+    'community_ressources' => ['new', 'show'],
+    'offers' => ['new', 'create'],
+    'public_pages' => ['about', 'contributions', 'landing_page', 'version'],
+    'thank_you' => ['show']
+  }
+
+  def self.shown_navbar?(controller, action, display_navbar)
+    always_display = NAVBAR_HIDEABLE[controller].nil? || !NAVBAR_HIDEABLE[controller].include?(action)
+    always_display || display_navbar
+  end
```
exbinary, line 14: One minor refactoring i might suggest:
```suggestion
  def self.shown?(controller, action, display_navbar)
    display_navbar || !hidden?(controller, action)
  end

  def self.hidden?(controller, action)
    NAVBAR_HIDEABLE[controller]&.include?(action)
  end 
```
* Since we're now using the method with the fully qualified module name, i think we don't need `navbar` in the method name
* Starting the conditional with `display_navbar ||` helps to clarify that the system system takes precedence over controller/action specific checks
* IMO, extracting the small `hidden?` method helps to simplify and make obvious the overall logic.

app/views/layouts/application.html.erb
```diff
@@ -3,7 +3,9 @@
   <%= render "layouts/head" %>
 
   <body>
-    <%= render "layouts/navbar" %>
+    <% if NavbarVisibility.shown_navbar?(controller_name, action_name, context.system_settings.display_navbar?) %>
+      <%= render "layouts/navbar" %>
+    <% end %>
```
exbinary, line 8: 💯

exbinary (approved):
Great adjustments, thank you! I had an idea for one small refactoring but it is completely optional and we can definitely merge as is.

app/lib/navbar_visibility.rb
gamecat2d: You are right ! Done.

h-m-m (approved):

Issue #755, opened 2020-09-29
Map view doesn't update as filters are changed


exbinary:

gamecat2d:
i give it a try !

exbinary:
Fixed in #970

PR #970, opened 2021-05-28
755 : Map view does not update

vurtn:
Sorry for my english.

### Why
The map view does not update when we change the filters (#755).

### What
- [x] Create a `geocode` method
- [x] Create an `add_marker` method
- [x] Create a watcher
- [x] Get the current state of the map 
- [x] Add a message if there is no match

### How
Ouch. So... The problem is that the component is never refreshed. The change of `contributions` is made but has no impact in the component. I changed several things:

First i splitted the function `loaded` to : `geocode` and `add_markers`. I extracted some variables in parameters to be able to output this logic if desired in a separate file. 

Then I realized that the markers are added when the component is created but we don't have access to them otherwise. The wrapper documentation didn't really help me so I fetched the map via `@map-init` to see what's available. It's the `mapbox-gl` map. So i searched how to remove the markers on it before drawing. In the method `add_markers` i remove the old points if they exists (`source` and `layer`) before adding the new ones. 

I check the contributions in the watcher to avoid the whole process if they are the same.

Instead of display an empty map if there is no match, i just printed a message. The result : 
map

### Testing
I don't know how to test this.

### Next Steps
Not seem to be concerned (or see the next point).

While trying to solve this issue, i wondered if it would not be good to change the library. I see 2 points to justify my point of view :
- The last update of `mapbox-gl-vue` is more than a year ago and the project seems abandoned. 
- I am not a lawyer so i'm maybe wrong but `mapbox-gl-vue` is `GPL-3.0 License`. On this [website](https://www.tawesoft.co.uk/kb/article/mit-license-faq) we can see : 
> Can I use GPL-licensed code in my MIT-licensed project?
> No. The project as a whole must conform to the terms of the GPL license and must be distributed under the terms of that license. Therefore such a project as a whole must be distributed as GPL, but can still contain MIT-licensed software. 

### Accessibility
Not seem to be concerned.

### Security
Not seem to be concerned.

### Meta
Not seem to be concerned.

### Pre-Merge Checklist

Best regards and stay safe !

### DIFF
--- a/app/javascript/pages/Browse.vue
+++ b/app/javascript/pages/Browse.vue
@@ -2,9 +2,8 @@
   <div>
     <div class="intro">
       <h1 class="title">Community Contributions</h1>
-      <p class="subtitle is-5">
-        Below are contributions (asks, offers, and community resources) that are available.
-        <br />You can filter through them by choosing filter options on the side.
+      <p class="subtitle is-5">Below are contributions (asks, offers, and community resources) that are available.
+        <br>You can filter through them by choosing filter options on the side.
       </p>
     </div>
 
@@ -19,12 +18,7 @@
       <section>
         <BrowserSelector :browser="browser" @clicked="browser = $event" />
       </section>
-      <component
-        :is="browser"
-        :contributions="activeContributions"
-        class="row"
-        accessToken="pk.eyJ1IjoibXV0dWFsLWFpZC1hcHAiLCJhIjoiY2tmZTBvd3UwMDBhbTJ4cDlic2JmMWZoaiJ9.rWscBjdl1SMT5N0yekIJYg"
-      />
+      <component :is="browser" :contributions="activeContributions" class="row" accessToken="pk.eyJ1IjoibXV0dWFsLWFpZC1hcHAiLCJhIjoiY2tmZTBvd3UwMDBhbTJ4cDlic2JmMWZoaiJ9.rWscBjdl1SMT5N0yekIJYg"/>
       <section class="has-text-centered mt-2">
         <h2 class="is-size-4 mb-1">Can't find what you're looking for?</h2>
         <div>
@@ -43,7 +37,6 @@ import ListBrowser from './browse/ListBrowser'
 import TileBrowser from './browse/TileBrowser'
 import MapBrowser from './browse/MapBrowser'
 import ContributionFetcher from './browse/ContributionFetcher'
-
 export default {
   components: {
     BrowserSelector,
@@ -54,14 +47,12 @@ export default {
   },
   props: {
     contributions: {type: Array},
-    filterGroupings: {type: Array, default: () => []},
+    filterGroupings: {type: Array, default: ()=>[]},
     initialFilters: {
       type: Array,
       default: function () {
         return [].concat(
-          ...this.filterGroupings.map((fGrouping) =>
-            fGrouping.filter_options.map((filter_option) => filter_option.id)
-          )
+          ...this.filterGroupings.map((fGrouping) => fGrouping.filter_options.map((filter_option) => filter_option.id))
         )
       },
     },
@@ -100,7 +91,7 @@ export default {
 </script>
 
 <style scoped>
-a {
-  margin: 0 0.5em;
-}
+  a {
+    margin: 0 0.5em;
+  }
 </style>


app/javascript/pages/browse/MapBrowser.vue
```diff
@@ -17,119 +17,169 @@
...
 
 <script>
-  import Mapbox from 'mapbox-gl-vue'
-  const mbxClient = require('@mapbox/mapbox-sdk');
-  const geoCoding = require('@mapbox/mapbox-sdk/services/geocoding');
-  //  FIXME Attempt to pass accessToken in script tag as seen in line 6
-  const baseClient = mbxClient({ accessToken: "pk.eyJ1IjoibXV0dWFsLWFpZC1hcHAiLCJhIjoiY2tmZTBvd3UwMDBhbTJ4cDlic2JmMWZoaiJ9.rWscBjdl1SMT5N0yekIJYg"});
-  const geoCodingService = geoCoding(baseClient)
+import Mapbox from 'mapbox-gl-vue'
+const mbxClient = require('@mapbox/mapbox-sdk')
```
h-m-m, line 32: I assume you've got an auto-formatter or something like that. Because it unindented all the code in this `script` section, we're having trouble seeing what did and did not change. Can you revert the whitespace changes so it's easier for us to review this PR?

h-m-m:
Thanks for digging into this @gamecat2d

We had one or two with Mapbox helping out Ruby for Good when this feature was first added. Let me see if we can contact them for help and suggestions about the points you raised, both about adding and removing markers and about the maintenance and licensing issues for the component.

It may take us a few weeks to get someone from Mapbox to respond. Hopefully it won't take too long

app/javascript/pages/browse/MapBrowser.vue
gamecat2d: Done :+1:

stanleypliu:
As a starting point to address @gamecat2d's point about switching libraries (I haven't worked with `mapbox-gl-vue` before), we could switch to using Mapbox GL JS as a CDN or NPM package instead, which is indeed supported (but making sure to use the right version as I believe v2 bills per map load).

marenab:
+1 for the switch to GL JS if that's an option. More documentation and support from Mapbox is possible if you're building with GL JS, like this one: https://docs.mapbox.com/mapbox-gl-js/example/filter-markers/

FWIW v1 also bills by map load (when using a Mapbox basemap), the difference with v2 is that it bills by map load regardless of where your basemap tiles come from. So there'd be no difference between v1 or v2 in terms of map loads billing if you're using a Mapbox basemap. Also, the Community team can help with donated/discounted services if you're concerned about surpassing the 50k free map loads per month: https://www.mapbox.com/community/

gamecat2d:
Sorry for my english.

Thanks @marenab and @stanleypliu  for your answers ! I have one more question : What is the advantage of mapbox over leaflet especially if the tiles are Open Street Map? The code seems easier to me with leaflet to do basic things and if you don't use mapbox tiles, you don't even need to have any API key. I discovered mapbox with this project and now i'm searching what are the possible alternatives. I don't understand the pros to use a commercial product (mapbox) for this project (mutual aid).

Have a great day and stay safe !

marenab:
That's a choice for your team to make @gamecat2d. Some folks prefer to build with Leaflet, some prefer GL JS. I'd say the main advantages of GL JS are that there are more supported features / documentation and the performance of vector tiles (instead of Leaflet's raster tiles) is smoother and faster to load. But your call! If you do choose Mapbox, the Mapbox Community team can support.

There is also a GL-Leaflet binding, if you'd like to build with Leaflet but bring in Mapbox vector tiles: https://github.com/mapbox/mapbox-gl-leaflet

h-m-m:
I've read over the comments in this thread and done some digging of my own.

@finn2d — I understand your interest in using other mapping tools. If I was doing this project by myself, I might have given them a much closer look before choosing. That said, we're doing this project with the help and support of the Ruby for Good community, which means we have a lot of people who can to give support us if we stick with Mapbox. (Like @marenab and @stanleypliu! Thanks you two for jumping in on this one.) So I'd prefer to stick with Mapbox for now, if that's okay. If that leads to problems later with API billing, I'm happy to wait to solve those problems once we have them.

I found that there are some more recent open source packages that tie Vue together to Mapbox, including the Mapbox GL JS tools. Since you have more experience in some of these mapping tools than I do, would you mind looking through what's available and seeing if any of them could solve our problems here while still sticking with Mapbox? Let me know what you think.

And again, thanks @finn2d for spending time on this. We really appreciate it!

Issue #882, opened 2021-02-26
Revisit CommunityResources index and their 'inexhaustibility'.


maebeale:

exbinary:
Addressed in #980 .

exbinary:
Mistakenly closed above. There's more work needed on this:
* Community Resources are now shown on `/contributions` but `/community_resources` still exists as a flat, unfilterable view. We should consider removing the latter or guiding users from there to the former.
* Need to revisit 'inexhaustibility' and how this is presented on `/contributions`.

Issue #880, opened 2021-02-26
Filter CommunityResources (use similar filter as on Contributions)


maebeale:

exbinary:
Superceded by #980.

PR #980, opened 2021-07-03
Spike of community resource browsing

h-m-m:
### Why

This addresses some of #882 and #880. Specifically, it 

- displays community resources in the contributions page
- allows community resources to be filtered, though only in the contributions page
- does mark the resources as inexhaustible 
- does not address claiming of community resources

### What
This updates our filtering classes so they can handle community resources.
The contact method logic is not implemented correctly, see inline notes below
There are certainly other aspects of this which are incomplete

Screen Shot 2021-07-03 at 11 07 56 AM

### Pre-Merge Checklist


app/models/offer.rb
```diff
@@ -6,6 +6,7 @@ class Offer < Listing
   scope :matched,   -> { includes(:matches_as_provider).references(:matches_as_provider).where.not(matches: {receiver_id: nil}) }
   scope :unmatched, -> { includes(:matches_as_provider).references(:matches_as_provider).where(matches: {receiver_id: nil}) }
   scope :matchable, -> { where(id: unmatched.pluck(:id) + inexhaustible.pluck(:id)) }
+  scope :in_service_areas, ->(ids) { where(service_area_id: ids) }
```
h-m-m, line 9: - [ ] TODO: refactor this to the parent class, assuming that doesn't cause any problems

app/models/community_resource.rb
```diff
@@ -31,20 +33,46 @@ def self.published
...
+
   def published?
     now = Time.current
     is_approved &&
       (publish_from.present? ? publish_from <= now : true) &&
       (publish_until.nil? || now < publish_until)
   end
 
+  def categories_for_tags
+    Category.where(name: tag_list)
```
h-m-m, line 47: Do we want to add validations? If you publish a community resource and it's missing either any service area or any category tag, it won't show up in the contributions page. So should we validate that if you're publishing a community resource it must have at least one service area and at least one category tag? Doing so makes a lot of sense to me. 

(The only alternative I can think of is adding a special case in the BrowseFilter logic — nor its partner in the Vue side — where it knows that something like "all categories are checked" means that category-less items can be returned. This alternative seems terrible to me as it's both a leaky abstraction and also I can think of a couple ways that it could go surprisingly wrong.)

app/models/community_resource.rb
```diff
@@ -31,20 +33,48 @@ def self.published
...
 
   def all_tags_to_s
     all_tags_unique.join(', ')
   end
+
+  def preferred_contact_method
+    # TODO: This is a hack that makes things work for now
+    # The test creates a contact method that will match this
+    # and the dev db seeding creates a couple, too
+    ContactMethod.method_name('call').last
```
h-m-m, line 62: Is there some way I can fix this in its own PR? In other codebases I work in, this whole workflow is something I'd hide behind a feature flag so I can get this PR out first, and then get the contact methods for community resources working in a separate PR. I could also try to fix it here, especially if anyone's willing to help me with it. Please let me know

app/blueprints/contribution_blueprint.rb
```diff
@@ -3,9 +3,12 @@
 class ContributionBlueprint < Blueprinter::Base
   identifier :id
   association :categories_for_tags, name: :category_tags, blueprint: DefaultBlueprint
-  association :service_area, blueprint: ServiceAreaBlueprint, view: :with_location
+  association :service_area, blueprint: ServiceAreaBlueprint, view: :with_location do |contribution|
+    contribution.service_areas.first
```
h-m-m, line 7: - [ ] TODO: @exbinary suggested that we could change this blueprint to output multiple service areas and the Vue code to display them. 
 
(Otherwise, we might have a community resource in areas `a` & `b` and someone filters by area `b` and is surprised to see a resource that shows up with only `a` listed as its area)

app/models/community_resource.rb
```diff
@@ -31,20 +33,48 @@ def self.published
...
+    "Community Resource"
+  end
+
+  def inexhaustible
+    true
+  end
+
+  def urgency_level_id
+    UrgencyLevel::TYPES.last.id
+  end
```
h-m-m, line 75: @exbinary mentioned that these don't feel like actual attributes, behaviors, or properties of a community resource, and so they might be better off in an abstraction or interface class that turns community resources into something blueprint-able

spec/blueprints/contribution_blueprint_spec.rb
```diff
@@ -73,4 +73,29 @@
...
+      "category_tags" => [],
+      "contact_types" => [{"id" => default_contact_method.id, "name" => "Call"}],
+      "contribution_type" => "Community Resource",
+      "created_at" => resource.created_at.to_f * 1000,
+      "description" => "Food for the revolution",
+      "inexhaustible" => true,
+      "location" => nil,
+      "match_path" => nil,
+      "name" => "Free breakfast for School Children Program",
+      "person" => nil,
```
h-m-m, line 92: @exbinary pointed out how hashes like this to test blobs of JSON are fragile since small changes can then require lots of edits

app/models/community_resource.rb
h-m-m: @svileshina mentioned that it's hard to assume we know what's right without talking to an organization about their specifics — and suggested perhaps a warning is better than an outright validation failure

h-m-m: I feel like http://c2.com/ppr/checks.html might be relevant here

app/models/community_resource.rb
h-m-m: Added an issue for this at https://github.com/rubyforgood/mutual-aid/issues/986

app/blueprints/contribution_blueprint.rb
h-m-m: Done!

app/models/community_resource.rb
h-m-m: Added validations with some comments in the commit

app/models/community_resource.rb
```diff
@@ -17,10 +17,17 @@ class CommunityResource < ApplicationRecord
 
   validates :name, :description, :publish_from, presence: true
 
+  # We're not sure if we need or want these validations. Adding them now assuming it will be easier to remove them later than the other way around
+  validates :service_areas, presence: true
+  validates_associated :service_areas
+  validates :tag_list, presence: true
```
h-m-m, line 23: - [ ] TODO: fix request specs that now flunk validation checks

PR #996, opened 2021-11-19
Make NavBarPolicy allow Feedback only for Admins

### LINKED ISSUE #992: Newly registered user is presented a Feedback option but not authorized to access it
Steps to recreate:
1. Register for a new account, confirm email
2. Click on the Feedback button in the nav bar

Actual:
User is shown the following message:
> Sorry, you aren't authorized to do that.
If you think this is wrong, please contact your administrator.

Expected:
User is either able to provide feedback or not shown the feedback button

H/T @vurtn for reporting in #963 .

bobmazanec:
### Why
To fix #992, show the `Feedback` NavBar button only to Admins & Sys-Admins

I chose those approach after seeing that the route triggered by the `Feedback` button uses the `SoftwareFeedbacksController` is derived from `AdminController`, which allows actions only by users with either the `admin` or `sys_admin` role. That is, I inferred the UI was incorrectly offering the button.

### What
Change `NavBarPolicy#visible_buttons` to include `Feedback` only for `User`s with either the `admin` or `sys_admin`.

### How
Moved adding `Feedback` to the allow-list from the line with `NavBarPolicy#visible_buttons`'s simple non-`nil` check to the line guarded with the call to `ApplicationPolicy#can_admin?`

### Testing
The PR modifies existing examples, removing the expectation for `Feedback`'s presence for `neighbor` cases.

#### With modified examples, without modified code - 2 failures
image

#### With modified examples & code - 0 failures
image

#### Rubocop - no change
image

More about the specs than the bug: in my dev/repro environment, I noticed that my new `User` had the `unset` role. This was different from `neighbor`, but both are "neither `admin` nor `sys_admin`", so I simply modified the existing `context`s.

### Accessibility
None that I'm aware of.

### Security
Very slight improvement—now the UI "advertises" one fewer route to non-{,sys}admin users.

### Meta

This is my first PR here.
I hope/believe I have read and understood the various guides, policies, etc. (humble apologies if not...)
Constructive feedback welcome, of course.

### Pre-Merge Checklist

### DIFF
--- a/app/policies/nav_bar_policy.rb
+++ b/app/policies/nav_bar_policy.rb
@@ -3,15 +3,8 @@ class NavBarPolicy < ApplicationPolicy
     visible = Set.new
     visible << (user ? 'Logout' : 'Login')
     visible << 'Sign Up' unless user
-    visible << 'Feedback' if is_admin?(user)
     visible.merge %w[Glossary Contributions] if system_settings.peer_to_peer?
-    visible.merge %w[Contributions Matches Admin] if can_admin?
+    visible.merge %w[Contributions Feedback Matches Admin] if can_admin?
     visible.to_a
-  end
-
-  private
-
-  def is_admin?(user)
-    user && (user.admin_role? || user.sys_admin_role?)
   end
 end
--- a/app/policies/nav_bar_policy.rb
+++ b/app/policies/nav_bar_policy.rb
@@ -3,9 +3,15 @@ class NavBarPolicy < ApplicationPolicy
     visible = Set.new
     visible << (user ? 'Logout' : 'Login')
     visible << 'Sign Up' unless user
-    visible << 'Feedback' if user
+    visible << 'Feedback' if is_admin?(user)
     visible.merge %w[Glossary Contributions] if system_settings.peer_to_peer?
     visible.merge %w[Contributions Matches Admin] if can_admin?
     visible.to_a
+  end
+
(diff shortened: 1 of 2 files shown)


app/policies/nav_bar_policy.rb
```diff
@@ -3,9 +3,15 @@ def visible_buttons
...
     visible.merge %w[Glossary Contributions] if system_settings.peer_to_peer?
     visible.merge %w[Contributions Matches Admin] if can_admin?
     visible.to_a
   end
+
+  private
+
+  def is_admin?(user)
+    user && (user.admin_role? || user.sys_admin_role?)
+  end
```
exbinary, line 16: The `ApplicationPolicy` base class provides a `can_admin?` method that serves the same purpose. In fact, looking at line 8 above, you can probably add `Feedback` to the array there instead.

exbinary (commented):
Thanks for working on this @bobmazanec ! 🙏🏾 .
Looks good, just one thought below.

app/policies/nav_bar_policy.rb
bobmazanec: Indeed—thanks for pointing that out (plus a little 🤦 😄)
Updated/simplified (e139258)

Anything else?

Thanks!

exbinary (approved):
Sweet, thanks for the quick turnaround!

